PR #3, opened 2020-08-31
readme: document a local test scenario

phlogistonjohn:
Add a local test scenario to the readme. Test is based on kubectl based
port forwarding and smbclient. These should work well for many Linux
desktop users.


obnoxxx (approved):
great, thanks!

PR #4, opened 2020-09-10
[WIP] Use operator framework v1

phlogistonjohn:
This switches the version of the operator framework from v0.17 to v1.0, using the guidelines here:
https://v1-0-x.sdk.operatorframework.io/docs/building-operators/golang/project_migration_guide/

Roughly speaking, the guide expects the project to be reinitialized and relevant code ported over from the previous version manually. This is what I have done so far. Due to the need to reinitialize I chose to base my work on the  commit *before* the older version of the operator framework was used. Thus, this PR loses a lot of the follow up work done on the makefile, helper scripts, etc. I do wish to bring much of that over.

Also, this has not been properly tested yet, only compiled. I am creating this WIP PR so you can comment on what has been done so far, and see incremental progress as I fill in the remaining gaps.

PR #5, opened 2020-09-27
Use operator sdk v1 - a cleaner approach

phlogistonjohn:
This switches the version of the operator framework from v0.17 to v1.0, using the guidelines here:
https://v1-0-x.sdk.operatorframework.io/docs/building-operators/golang/project_migration_guide/

Roughly speaking, the guide expects the project to be reinitialized and relevant code ported over from the previous version manually. This is roughly what I have done so far. Due to the need to reinitialize I first refactored a bunch of the logic inside the controller package created by the previous version of the sdk. Then I had to delete files that the operator sdk "takes over". After that I performed the porting steps in the guide, using the new refactored types to minimized the code added to the new controller package.

There are a few small additional tweaks to the code and the supporting files, largely to tix issues that prevented me from testing by running through the steps outlined in the readme, while at the same time updating the readme as appropriate. Otherwise, the functionality should be equivalent to the previous version.


phlogistonjohn:
I am leaving PR #4 open for now but I do believe that this PR obsoletes that one. I am leaving open as a point of comparison. Some of these patches are cherry picks from that work, but others are wholly original to this branch.

Issue #2, opened 2020-08-28
Why two CRD types?


phlogistonjohn:
I was trying out this operator and following along with the readme but couldn't figure out the justfication for the two different CRDs. It seems like the SmbPvc is meant to create a SmbService and a PVC and glue them together. If I'm right I'm not quite sure what the advantage is.

If I may, my first thought is that the CRDs should be more oriented toward the task the user (cluster/storage admins) are going to be performing. In that way, SmbService makes some sense but you perhaps even go more granluar and create SmbShare CRDs instead.

IMO if you want to bind the lifecycle of a PV/PVC from a lower part of the storage stack either by name, for an existing PVC, or by directly specifying the pvc parameters I'd go with something like:
```yaml
source:
  pvc:
    name: "mypvcname"
```
AND as an alternate form
```yaml
source:
  pvc:
    spec:
        #...embedded pvc spec...
```

phlogistonjohn:
I forgot to mention that setting name and spec under pvc are mutually exclusive.

phlogistonjohn:
The next phase CRD design unifies share level types under one CRD. This work has started. This can be closed once the SmbService and SmbPvc types are removed.

phlogistonjohn:
There's no reason to keep this open now.

PR #6, opened 2020-10-07
RFC: unified api type - supporting pvc name or spec

phlogistonjohn:
The following changes are based on PR #5 - those patches are part of this PR at the moment.

This is an (incomplete) demonstration of an alternative approach to having two CRDs, something I questioned in issue #2.

The good:
* One CRD type to document and for users to understand
* Any logic where SmbPvc needs to manage SmbServices will just go away, reducing the total number of types being managed
* Operator only needs to  monitor one type

The bad:
* There doesn't seem to be a simple way to make two mutually exclusive options at kube submission time. The code that checks for this only happens in reconcile. To make this more elegant for the user I think an admission webhook would be needed to reject json/yamls that try to set both pvc name and pvc spec (or leave both out).

If you think the good outweighs the bad, as I still do, and I will continue this effort and remove the SmbPvc type. After that we can discuss naming. :-)


phlogistonjohn:
obsoleted by PR #19  - closing.

PR #7, opened 2020-10-09
examples: adapt yamls to new namespace

obnoxxx:
Adapt examples for new namespace `samba-operator.samba.org` instead of `smbservice.samba.org` and `smbpvc.sam


phlogistonjohn (approved):
+1. Missed those files previously.

PR #13, opened 2020-11-02
makefile: try to force use of shell

phlogistonjohn:
Currently the CI is not properly handling command -v, as make bypasses
the shell.
See also:
https://stackoverflow.com/questions/5618615/check-if-a-program-exists-from-a-makefile


obnoxxx:
this was included in #14

PR #14, opened 2020-11-02
Rework revive detection and installation

obnoxxx:
- include the trick from #13 to make gnu make invoke the shell by adding `;`
- first check revive in path, then in GOBIN, as last resort, download.
- add informational messages (can be skipped if not desired)
- make revive PHONY

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,17 +148,22 @@ check-revive: revive
 
 revive:
 ifeq (, $(shell command -v revive ;))
+	@echo "revive not found in PATH, checking in GOBIN ($(GOBIN))..."
 ifeq (, $(shell command -v $(GOBIN)/revive ;))
 	@{ \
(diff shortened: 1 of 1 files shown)


phlogistonjohn (approved):
Seems OK to me. It doesn't work for me if I remove revive from my PATH because for me, GOBIN is not on my path. Regardless this is overall an improvement and I can mess with it some more later to be smarter about PATH.

obnoxxx:
> Seems OK to me. It doesn't work for me if I remove revive from my PATH because for me, GOBIN is not on my path. Regardless this is overall an improvement and I can mess with it some more later to be smarter about PATH.

It should work if GOBIN is not in your PATH. It works for me at least.

obnoxxx:
merging after confirmation by @phlogistonjohn

Issue #10, opened 2020-10-21
Drop port 139


phlogistonjohn:
I've been researching how best to expose SMB shares to systems outside the k8s cluster. One thing that stands out to me is that we really don't need port 139. It's pretty old, obsolete, and having it as part of the container spec is just going to be confusing. I think we should focus our efforts on "modern" SMB until we have strong demand otherwise.

obnoxxx:
Agreed

PR #15, opened 2020-11-03
resources: don't expose 139 as a container port

phlogistonjohn:
Port 139 is fairly obsolete and we shouldn't needed for our
use cases.

Fixes: #10


obnoxxx (approved):
lgtm

PR #18, opened 2020-11-05
resources: extend and reorganize the set of labels applied by smbservice

phlogistonjohn:
When the smbservice is used to create deployments, etc. this more
comprehensive set of labels will be applied.


obnoxxx (approved):
@phlogistonjohn - I would usually request the commit msg and/or PR description to mention *why* this is done / what is the benefit of doing it. But just accepting this now.

PR #19, opened 2020-11-09
docs: add a next-phase CRD design document

phlogistonjohn:
I've been working on this in the background and I think it's good enough to share now. I do not expect to merge this as-is but to mainly use it to kick off additional discussion. That said, I think it would be good to merge it once we agree on the broad strokes and then continue to modify it there.

Please, do ask questions and make me justify some of these ideas! :-)

Part of: #2


docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,215 @@
...
+* SmbSecurityDomain - A CR that encapsulates the knowledge needed to create "local" users or bind to an Active Directory domain
+
+The operator will take the SmbShare and SmbSecurityDomain resources as inputs
+and create as many smbd, winbind, etc. backing services as needed. Users
+will have limited input into what backing services the operator will create.
+The operator may, or may not, combine one or more share into a single smbd
+instance.
+
+SmbShare CRs will refer to one or more SmbSecurityDomain resources. The
+SmbSecurityDomain refernce will define the main security properties of the
```
obnoxxx, line 28: `refernce` ==> `reference`

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,215 @@
...
+  * `backend` - enumerated string - the name of a (supported) ID mapping backend
+  * `domainName` - string - the name of the domain being configured
+  * TBD - options to configure the backend, ID range, etc.
+
+
+## SmbShare
+
+Spec Options:
+* `shareName` - string - Optional string giving the share a full SMB compliant name. If unset, the name will be derived by the operator.
+* `storage` - mapping - A new subsection to configure the storage backing the share.  * `pvc` - mapping - A new subsection used to configure PVC backing storage.
```
obnoxxx, line 57: newline missing

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,215 @@
...
+    * `name` - string - The name of a PVC, managed externally to the operator, for the share
+    * `spec` - mapping, embedded pvc spec - An embedded PVC spec that will be used to dynamically create a backing PVC for the share; sharing the lifecycle of the PVC with the share
+  * TBD - Any other more custom storage backends if needed
+* `securityDomain`  - string - The name of the SmbSecurityDomain CR associated with this share
+* `grouping` - string - Optional string that if undefined will be generated by the operator. The grouping is used by the operator as a hint to combine shares under one smbd service, depending on other settings
```
obnoxxx, line 62: What do you imagine the value of `grouping`to look like? A list of other smbshare names?

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,215 @@
...
+* `clustering` - mapping - A new subsection pertaining to clustering smbd instances
+  * `mode` - enumerated string "none" or "ctdb" - This option tells the operator if this share should be clustered or not
+  * `minSize` - int - Minimum number of smbd instances when clustering
```
obnoxxx, line 65: Why minimum and not an exact size?

docs/design/crd-proposal-phase1.md
obnoxxx: OK, I see it in the examples below. Just a group name that will have all the shares with the same grouping name combined under on smbd instance.

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,215 @@
...
+
+Spec Options:
+* `shareName` - string - Optional string giving the share a full SMB compliant name. If unset, the name will be derived by the operator.
+* `storage` - mapping - A new subsection to configure the storage backing the share.  * `pvc` - mapping - A new subsection used to configure PVC backing storage.
+    * `name` - string - The name of a PVC, managed externally to the operator, for the share
+    * `spec` - mapping, embedded pvc spec - An embedded PVC spec that will be used to dynamically create a backing PVC for the share; sharing the lifecycle of the PVC with the share
+  * TBD - Any other more custom storage backends if needed
+* `securityDomain`  - string - The name of the SmbSecurityDomain CR associated with this share
+* `grouping` - string - Optional string that if undefined will be generated by the operator. The grouping is used by the operator as a hint to combine shares under one smbd service, depending on other settings
+* `clustering` - mapping - A new subsection pertaining to clustering smbd instances
```
obnoxxx, line 63: We might want to structure this a little differently, not yet sure how: As mentioned I was identifying 3 aspects of samba's dbwrap backend currently provided by ctdb
1. indication of cluster intactness in the backend (recovery lock file)
2. volatile DBs (and messaging)
3. persistent DBs.

of these, #1 might be useful without clustering, #3 is definitely useful for making config and settings (join...) persistent. But #2 would really for moving to all-active clustering. This would require development in samba though. 

So to summarize, not sure if putting this under "clustering" is entirely sufficient or appropriate.

obnoxxx (commented):
In general, I like the idea of structuring it this way very much. A couple of comments and thoughts inside.

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,215 @@
...
+Spec Options:
+* `mode` - enumerated string "user" or "active-directory" - Indicates that this is an AD joined instance or based on a local mapping of users
+* `users` - mapping - A new subsection pertaining to "user" options - required when mode = "user"
+  * `secret` - string - The name of a secret containing user and group definitions that will be used to create the local users and groups for the service
+* `localName` - string - Optional string used to indicate how this instance will
+  be referred to in an Active Directory
+* `realm` - string - The dns domain name of an Active Directory domain
+* `domains` - list-of-maps - A new subsection pertaining to detailed "active-directory" configuration (optional)
+  * `backend` - enumerated string - the name of a (supported) ID mapping backend
+  * `domainName` - string - the name of the domain being configured
```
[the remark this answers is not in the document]
phlogistonjohn: The `realm` translates to the smb.conf parameter of the same name. The `domainName` keys loosely map to the domain key of the `idmap` smb.conf settings. These are optional but allow one to say configure a non-default idmap setting like "ad" for either the primary `realm` or other trusted domains.

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,215 @@
...
+## SmbSecurityDomain
+
+Spec Options:
+* `mode` - enumerated string "user" or "active-directory" - Indicates that this is an AD joined instance or based on a local mapping of users
+* `users` - mapping - A new subsection pertaining to "user" options - required when mode = "user"
+  * `secret` - string - The name of a secret containing user and group definitions that will be used to create the local users and groups for the service
+* `localName` - string - Optional string used to indicate how this instance will
+  be referred to in an Active Directory
+* `realm` - string - The dns domain name of an Active Directory domain
+* `domains` - list-of-maps - A new subsection pertaining to detailed "active-directory" configuration (optional)
```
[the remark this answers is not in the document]
phlogistonjohn: It is supposed to contain as many settings for domains as you have (primary or trusted) that you want to configure.

Example:
```
domains:
  - domainName: myrealm.mycorp.example.com
    backend: ad
  - domainName: atrust.mycorp.example.com
    backend: ad
```

Do you you think it would be clearer as a map-of-maps like this example:
```
domains:
    myrealm.mycorp.example.com:
        backend: ad
    atrust.mycorp.example.com:
        backend: ad
```?

If we need to the section could include other parameters such as idmap ranges, etc. But I left it out of the initial draft becase (a) I was hoping to keep this siimple and (b) hoping the operator could do the heavy lifting of determining valid ranges after being simply told to use the ad (and check for rfc2307 in the ad).

File: docs/design/crd-proposal-phase1.md [hunk 8]
```diff
@@ -0,0 +1,215 @@
...
+* SmbShare - A CR that encapsulates all the information needed to export a single share
+* SmbSecurityDomain - A CR that encapsulates the knowledge needed to create "local" users or bind to an Active Directory domain
+
+The operator will take the SmbShare and SmbSecurityDomain resources as inputs
+and create as many smbd, winbind, etc. backing services as needed. Users
+will have limited input into what backing services the operator will create.
+The operator may, or may not, combine one or more share into a single smbd
+instance.
+
+SmbShare CRs will refer to one or more SmbSecurityDomain resources. The
```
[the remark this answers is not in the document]
phlogistonjohn: It would not. This is poorly worded. I will improve it.

docs/design/crd-proposal-phase1.md
phlogistonjohn: I wasn't sure if we'd want to have a minSize and a maxSize in the future. But perhaps that is premature. Would you prefer if I changed it to just "size" and we can deal with the future in the future?

docs/design/crd-proposal-phase1.md
phlogistonjohn: Sure. At the time I was only thinking of non-clustered or clustered. I do think that some sort of subsection would be needed to what kind of deployment (in the general sense, not just the k8s resource) we want and how it is expected to scale. Perhaps if we replace "clustering" with "scaling" as a more generic term? And then the type of scaling might be single instance or clustered or some hybrid...

phlogistonjohn: Also, I forgot to note. None of this is discussing what the defaults would be unless explicitly stated. I suspect that in some cases we might prefer to have the more "complex" mode the default but allow some use cases to "opt-out". For example if a clustered mode was the default, a user might want to scale down to one instance for a read only share of limited importance to save on resources, etc.

docs/design/crd-proposal-phase1.md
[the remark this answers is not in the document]
phlogistonjohn: updated.

docs/design/crd-proposal-phase1.md
phlogistonjohn: fixed.

docs/design/crd-proposal-phase1.md
phlogistonjohn: should be fixed.

phlogistonjohn:
Pushed an updated version that fixes some of the items that didn't warrant additional discussion. I'm looking forward to hearing your opinions on the remaining items!

docs/design/crd-proposal-phase1.md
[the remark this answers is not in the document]
obnoxxx: I think a tag for the name is fine, but `domainName` under `domains` is kind of redundant. we could just have `name` under `domains`?

docs/design/crd-proposal-phase1.md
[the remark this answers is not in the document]
obnoxxx: Got it.

obnoxxx (commented):
After your comments and thinking about this a bit more, I think we should rename `SmbSecurityDomain`: This fundamentally corresponds to the global configuration of a samba server (`smb.conf` file). The `SmbShare` type corresponds to a share section in an `smb.conf` file. 

So in this regard, what you named `SmbSecurityDomain` might better be named `SambaInstance` or `SambaServer`. The instance is ultimately identified by its internal ID (the machine SID) once the server/db is initialised. The fqdn (essentially `localName` + `realm`) can also identify the instance.

There can only be one main smbd for a given SID / fqdn. So the `SmbShare` CRs need to end up as share config sections for their respective smbd instances.

Each `SmbShare` needs to be associated to one and only one `SambaServer` (or `SambaSecurityDomain`). Also very important that (1) each PVC is only attached to only one `SambaServer` (indirectly). If there is a possibility for a subdirectory of a PVC to be turned into another PVC, then it is important that these two PVCs are attached to the *same* `SambaServer`.

docs/design/crd-proposal-phase1.md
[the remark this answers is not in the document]
phlogistonjohn: Sure. I thought a more descriptive key would be better but it is somewhat redundant.

phlogistonjohn:
> After your comments and thinking about this a bit more, I think we should rename `SmbSecurityDomain`: This fundamentally corresponds to the global configuration of a samba server (`smb.conf` file). The `SmbShare` type corresponds to a share section in an `smb.conf` file.
> 

I agree that the security & domain settings reside in the global section of the smb.conf. I also agree that SmbShare aligns with the share section of smb.conf.


> So in this regard, what you named `SmbSecurityDomain` might better be named `SambaInstance` or `SambaServer`. The instance is ultimately identified by its internal ID (the machine SID) once the server/db is initialised. The fqdn (essentially `localName` + `realm`) can also identify the instance.
> 

I was trying to explicitly avoid tying the operator's config to either server instances or smb.conf. The names of the many config items reflect the parameters in smb.conf my intention is not to pass these options to smb.conf in a  1:1 manner. Rather the names are used for familiarity.  

I prefer SmbSecurityDomain because:
* It only covers the security and domain settings, while real smb.conf has many many more global options
* It emphasizes the important user provided nature of security settings
* It manages some important security and domain settings that typically lie outside of smb.conf - mainly the reference to a secret holding user information for non-domain-joined mode


> There can only be one main smbd for a given SID / fqdn. So the `SmbShare` CRs need to end up as share config sections for their respective smbd instances.
> 

Yes.

> Each `SmbShare` needs to be associated to one and only one `SambaServer` (or `SambaSecurityDomain`). 

Agree, when talking about a non-clustered scenario. For clustering its a tad bit ambiguous how to describe this as I assume you see the mutliple clustered smbd as sharing one config, right?

> Also very important that (1) each PVC is only attached to only one `SambaServer` (indirectly). If there is a possibility for a subdirectory of a PVC to be turned into another PVC, then it is important that these two PVCs are attached to the _same_ `SambaServer`.

Again, I'm a little unclear on how you think about this in a clustered situation. I envision one Read-Write-Many PVC would be attached to as many pods as needed. I agree that this PVC would be coupled to one (SmbShare, SmbSecurityDoamin) instance, however many smbds/pods that may entail.

docs/design/crd-proposal-phase1.md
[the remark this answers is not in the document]
phlogistonjohn: changed.

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,220 @@
...
+create shares that can be consumed by a variety of clients for various use
+cases. Some sections are deliberately marked "TBD" (to be decided) as
+placeholders for more detail that will be needed even in Phase 1.
+
+## SmbSecurityDomain
+
+Spec Options:
+* `mode` - enumerated string "user" or "active-directory" - Indicates that this is an AD joined instance or based on a local mapping of users
+* `users` - mapping - A new subsection pertaining to "user" options - required when mode = "user"
+  * `secret` - string - The name of a secret containing user and group definitions that will be used to create the local users and groups for the service
```
gd, line 45: What exactly does the 'secret' string refer to? Is it some sort of embedding user and group definitions in a  cryptographical manner?

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,220 @@
...
+```yaml
+---
+apiVersion: samba-operator.samba.org/v1alpha1
+kind: SmbSecurityDomain
+metadata:
+  name: "local-users"
+spec:
+  mode: "user"
+  users:
+    secret: "local-users-secret"
```
gd, line 88: Just curious: What is that secret used for?

docs/design/crd-proposal-phase1.md
phlogistonjohn: In kubernets a secret is a bundle of configuration that may (or may not) be protected by encyption. By default it is simply obscured but you can install add ons that can protect the secret contents more thoroughly.

docs/design/crd-proposal-phase1.md
phlogistonjohn: I envision it would contain a json configuration of local users and groups, the actual structure of said json TBD of course. :-)

phlogistonjohn:
After a verbal discussion I came to the realization that 'localName" was not appropriate for a CR that would be usable by multiple instances. I have renamed the field "instanceNamePrefix" and updated the description.

phlogistonjohn:
As discussed verbally I'd be OK with renaming "SmbSecurityDomain" to "SmbIdentity". It's shorter and is as meaningful as SmbSecurityDomain IMO. Let me know what you think.

docs/design/crd-proposal-phase1.md
obnoxxx: Note from earlier discussion: minSize was chosen to allow for more flexible grouping...

obnoxxx:
> As discussed verbally I'd be OK with renaming "SmbSecurityDomain" to "SmbIdentity". It's shorter and is as meaningful as SmbSecurityDomain IMO. Let me know what you think.

I think that `SmbIdentity` would really mean that we'd have a 1:1 correspondence between this and an instance. An "Identity" could conceptually by its name not serve as a template for instances.

Maybe call it `SambaSecurityConfig` or `SambaSecuritySetup` instead?

@phlogistonjohn

phlogistonjohn:
> > As discussed verbally I'd be OK with renaming "SmbSecurityDomain" to "SmbIdentity". It's shorter and is as meaningful as SmbSecurityDomain IMO. Let me know what you think.
> 
> I think that `SmbIdentity` would really mean that we'd have a 1:1 correspondence between this and an instance. An "Identity" could conceptually by its name not serve as a template for instances.
> 
> Maybe call it `SambaSecurityConfig` or `SambaSecuritySetup` instead?
> 

I'd go with SmbSecurityConfig to match the other CR name. But I'm fine with ...Config as the suffix rather than "Domain".

obnoxxx:
> I'd go with `SmbSecurityConfig` to match the other CR name. But I'm fine with ...Config as the suffix rather than "Domain".

Oh, sure, I mistyped Samba for Smb, `SmbSecurityConfig` is good!

phlogistonjohn:
I have pushed a new version. 
It changes SmbSecurityDomain to SmbSecurityConfig.
All options pertaining to clustering have been moved under "scaling" and the names are tweaked but  are functionally the same.
Grammar and spelling improvements.
Block formatting on some of the long lines.

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,240 @@
...
+  storage:
+    pvc:
+      name: "docs"
+---
+apiVersion: samba-operator.samba.org/v1alpha1
+kind: SmbShare
+metadata:
+  name: "cad-files"
+spec:
+  securityDomain: "local-users"
```
obnoxxx, line 135: `securityConfig`

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,240 @@
...
+      name: "cool-ad.int.example.com"
+    - backend: ad
+      name: "my-trust.int.example.com"
+---
+apiVersion: samba-operator.samba.org/v1alpha1
+kind: SmbShare
+metadata:
+  name: "documents"
+spec:
+  securityDomain: "cool-domain"
```
obnoxxx, line 166: `securityConfig`

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,240 @@
...
+  storage:
+    pvc:
+      name: "docs"
+---
+apiVersion: samba-operator.samba.org/v1alpha1
+kind: SmbShare
+metadata:
+  name: "cad-files"
+spec:
+  securityDomain: "cool-domain"
```
obnoxxx, line 176: `securityConfig`

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,240 @@
...
+Two shares, with groups:
+```yaml
+# SmbSecurityConfig not shown
+---
+apiVersion: samba-operator.samba.org/v1alpha1
+kind: SmbShare
+metadata:
+  name: "documents"
+spec:
+  securityDomain: "local-users"
```
obnoxxx, line 194: `securityConfig`

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,240 @@
...
+  storage:
+    pvc:
+      name: "docs"
+---
+apiVersion: samba-operator.samba.org/v1alpha1
+kind: SmbShare
+metadata:
+  name: "cad-files"
+spec:
+  securityDomain: "local-users"
```
obnoxxx, line 207: `securityConfig`

docs/design/crd-proposal-phase1.md
```diff
@@ -0,0 +1,240 @@
...
+One share with clustering enabled:
+```yaml
+# SmbSecurityConfig not shown
+---
+apiVersion: samba-operator.samba.org/v1alpha1
+kind: SmbShare
+metadata:
+  name: "documents"
+spec:
+  securityDomain: "cool-domain"
```
obnoxxx, line 228: `securityConfig`

docs/design/crd-proposal-phase1.md
obnoxxx:
```suggestion
  securityConfig: "local-users"
```

docs/design/crd-proposal-phase1.md
obnoxxx:
```suggestion
  securityConfig: "cool-domain"
```

docs/design/crd-proposal-phase1.md
obnoxxx:
```suggestion
  securityConfig: "cool-domain"
```

docs/design/crd-proposal-phase1.md
obnoxxx:
```suggestion
  securityConfig: "local-users"
```

docs/design/crd-proposal-phase1.md
obnoxxx:
```suggestion
  securityConfig: "local-users"
```

docs/design/crd-proposal-phase1.md
obnoxxx:
```suggestion
  securityConfig: "cool-domain"
```

obnoxxx (changes requested):
Looks good in general.
A few cases in the examples with the old names.

obnoxxx (commented):
Oh, and is the duplicate signoff in the commit message intentional?

phlogistonjohn:
> Oh, and is the duplicate signoff in the commit message intentional?

Heh. No. It's a side effect of something I am experimenting with.

obnoxxx (approved):
lgtm

PR #20, opened 2020-11-22
Fix revive detection

obnoxxx:
There was an error in revive detection. If revive was installed by make, the variable REVIVE was not set correctly due to simple variable assignment. Recursive assignment fixes this.


phlogistonjohn (approved):
Usually, the recursive mode is the cause of issues. Happy to see it fix an issue for once. ;-)

PR #25, opened 2021-02-05
Add the initial support for the SmbShare (pt. 1 of many)

phlogistonjohn:
It's a tad bit messy, but trying to move from SmbService and SmbPvc to SmbShare piece by piece.

These changes add SmbShare and then add the bare minimal set of features to make it feature equivalent to SmbService and SmbPvc.
Future changes should remove SmbService and SmbPvc and then begin expanding on SmbShare to more closely match the design doc.

It's hand tested.


internal/resources/smbshare.go
```diff
@@ -0,0 +1,217 @@
...
+	err := m.client.Get(ctx, nsname, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return Done
+		}
+		// Error reading the object - requeue the request.
+		return Done
```
obnoxxx, line 62: Not sure if this is what you intended, but comment and code don't seem to match.

```suggestion
		return Requeue
```

obnoxxx (changes requested):
Looks good in general. One small detail caught my attention. See inline.

internal/resources/smbshare.go
phlogistonjohn: You are right. This mistake is present in the other CR managers and I copy and pasted it here. I have a fix in a change for a future PR. If you like, we can fix it in this PR, or we can ignore it temporarily and take the fix in a follow up. Let me know how you feel.

obnoxxx: why not get it right when we add it?

obnoxxx (approved):
Thanks for the update! LGTM

PR #27, opened 2021-02-07
github actions: test container build on PRs

obnoxxx:
Signed-off-by: Michael Adam <obnox@redhat.com>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,6 +36,8 @@ jobs:
     if: github.event_name == 'pull_request'
     steps:
       - uses: actions/checkout@v2
+      - name: check free space
+        run: df -h
       - name: build container image
         # note: forcing use of podman here since we are
         # using podman explicity for the push job
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -63,10 +63,12 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: log in to quay.io
-        run: podman login -u "${{ secrets.QUAY_USER }}" -p "${{ secrets.QUAY_PASS }}" quay.io
+        # using docker for now, since podman has an issue with space
+        # consumption: image build fails with no space left on device...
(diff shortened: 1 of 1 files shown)


obnoxxx:
After merging PR #26, I noticed that the push action failed on the container build with `no space left on device`.
Therefore, I am adding the container build to the PR tests as well (not the push though...).

obnoxxx:
OK, this doesn't really pass. I just added an ` || df -h` after the make command.
It fails with no space left, but the df does not seem to confirm that... 🤷

obnoxxx:
OK... Added a docker build flavour, and that passes...
Not sure what's up with podman and space consumption.

obnoxxx:
OK, so ...
* container build using docker works
* container build using podman fails due to no space left on device

@phlogistonjohn , should we just use docker for now, and have a PR check job for both podman and docker?

obnoxxx:
@phlogistonjohn, I changed this to:

1. build with both podman and docker in PR verification (podman currently fails)
2. change the push job to use docker (which should work for now)

phlogistonjohn:
> @phlogistonjohn, I changed this to:
> 
>     1. build with both podman and docker in PR verification (podman currently fails)
> 
>     2. change the push job to use docker (which should work for now)

Sounds good.
As for the podman issue itself we should check what version these builder vms come with. We could also see if we can pull in a newer release if its old, and see if that improves things. If not, perhaps we should report the issue.

obnoxxx:
> > @phlogistonjohn, I changed this to:
> > ```
> > 1. build with both podman and docker in PR verification (podman currently fails)
> > 
> > 2. change the push job to use docker (which should work for now)
> > ```
> 
> Sounds good.

Should we push it as is then and let the podman build fail for now?

> As for the podman issue itself we should check what version these builder vms come with. We could also see if we can pull in a newer release if its old, and see if that improves things. If not, perhaps we should report the issue.

https://github.com/actions/virtual-environments/issues/709
https://github.com/actions/virtual-environments/issues/1860
https://github.com/containers/podman/issues/3846

phlogistonjohn (approved):

PR #28, opened 2021-02-08
Make make test work

obnoxxx:
Fix make test to be autonomous according to https://github.com/operator-framework/operator-sdk/pull/3983 .
And reformat it a bit.
Add "make test" to the CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,6 +29,14 @@ jobs:
     - name: Run checks
       run: make check
 
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/setup-go@v2
+    - uses: actions/checkout@v2
+    - name: run the tests
+      run: make test
+
   podmanbuild:
     runs-on: ubuntu-latest
     # don't run on push, since the "push" job contains the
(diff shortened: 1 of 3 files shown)


obnoxxx:
Oh, this fails. Weird:

```
/bin/sh: 1: source: not found
/bin/sh: 2: fetch_envtest_tools: not found
/bin/sh: 3: setup_envtest_env: not found
```

Guess that's because running on ubuntu (`dash` as sh...)

obnoxxx:
Next bashes, I guess:
```
/bin/sh: 18: set: Illegal option -o pipefail
	go test ./... -coverprofile cover.out
```

phlogistonjohn (changes requested):
The CI has failed. There seem to be more "bashisms" present.

I think it would possibly be better to remove all of the "excess" shell scripting from the makefile and put it in a (bash) script inside of the "hack" directory.

obnoxxx:
> The CI has failed. There seem to be more "bashisms" present.

Yes, as I mentioned in the previous comment. 😉 

> I think it would possibly be better to remove all of the "excess" shell scripting from the makefile and put it in a (bash) script inside of the "hack" directory.

Probably the best solution....

Now I understand the problem: The downloaded `setup-envtest.sh` script starts with `#!/usr/bin/env bash` but it is sourced instead of executed, so the shebang is not useful... That is basically broken in the sdk itself. 

Will just put it into a script as suggested...

obnoxxx:
Yay, `test` target passed now! 😄

phlogistonjohn (approved):
I don't love how much this thing downloads stuff when run, but it's better than not working. We can always improve it more later.

PR #29, opened 2021-02-08
github actions: provide password via stdin

obnoxxx:
This is more secure than providing the password as a command line
argument to docker login.

Signed-off-by: Michael Adam <obnox@redhat.com>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -73,7 +73,7 @@ jobs:
       - name: log in to quay.io
         # using docker for now, since podman has an issue with space
         # consumption: image build fails with no space left on device...
-        run: docker login -u "${{ secrets.QUAY_USER }}" -p "${{ secrets.QUAY_PASS }}" quay.io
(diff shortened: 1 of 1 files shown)


phlogistonjohn (approved):
If it stops docker from complaining its worth it.

obnoxxx:
> If it stops docker from complaining its worth it.

heh, it seems it didn't:

```
Run echo "***" | docker login -u "***" --password-stdin quay.io
WARNING! Your password will be stored unencrypted in /home/runner/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store
Login Succeeded
```

So I must have mis-interpreted the warning... 🤦

phlogistonjohn:
:man_shrugging: It can't be worse than before.

We could look into using robot logins for github.

I know quay io has this text on my account:
```
The Docker CLI stores passwords entered on the command line in plaintext. It is therefore highly recommended to generate an an encrypted version of your password to use for docker login. 
```
If we haven't already done that we can generate a "robot login" for the quay io repo.

obnoxxx:
I created an encrypted password for this purpose. It is not using the clear text pwd.

PR #31, opened 2021-02-16
Replace the old SmbService and SmbPvc types (pt. 2 of many) 

phlogistonjohn:
PR #25 added the new SmbShare type. This PR performs the next major step by making SmbService (roughly) feature equivalent with the old types and then removing those types.


obnoxxx (approved):
lgtm

PR #33, opened 2021-02-17
api: set proper defaults for our share booleans

phlogistonjohn:
Use the "annotations" provided by kubebuilder to specify how to treat the boolean values when they're not explicitly set.


phlogistonjohn:
This isn't as standalone as I thought. This needs to make use of `apiextensions.k8s.io/v1` and because of that probably needs to be rebased on the patches in PR #31.

phlogistonjohn:
This is now based on PR #31 . This was done because setting the controller-gen to use `apiextensions.k8s.io/v1` would side effect all the crd files, including the ones were in the process of deleting and create a bunch of churn and merge conflicts.
Please review PR #31 first. Sorry for the noise. :-)

obnoxxx (approved):
lgtm

PR #35, opened 2021-02-23
Introduce share planner and configmap

phlogistonjohn:
The configmap contains a json based configuration file that is consumed by the sambacc library. This is used to initialize the container and configure samba. The share planner is a new type created to manage "knowledge" about this configuration and the exceptions about containerized samba w/o directly interacting with kubernetes constructs.

Some logging is also added as the number of steps the update function takes is begging to increase.

Based on PR #33 . Please review that first.

### DIFF
--- a/internal/resources/configmap.go
+++ b/internal/resources/configmap.go
@@ -1,0 +1,106 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


obnoxxx (changes requested):
Check failing needs to be addressed. Otherwise - looks good.

obnoxxx (approved):
lgtm

PR #38, opened 2021-03-02
Add new CRD for SmbSecurityConfig

phlogistonjohn:
These patches add the planned SmbSecurityConfig CR. It is not a full implementation of all the features noted in the design but rather a bare minimum subset. However, I did update the design doc to add a key field to the user and groups secret as it was missing.


obnoxxx (approved):
lgtm

PR #39, opened 2021-03-02
Add basic Active Directory domain support to SmbSecurityConfig

phlogistonjohn:
Depends on PR #38. Please review that first.

These patches build on the recently added SmbSecurityConfig to add support for supporting AD enabled shares by running samba as an Active Directory domain member. It supports a basic method for doing joins, or if not enabled the pods rely on the sambacc tool to "pause" the must-join pod until a human runs `samba-container join` in the pod (via kubectl exec or the like).
The design doc lacked a way to specify how to join and where to source the join data, so that doc has been updated.

The ability to configure individual domains and trusts is present but is in a very primitive state and we'll probably want to work on this area in the future for both UX/workflow and implementation.


obnoxxx:
For some reason, merging #38 has created conflicts here...

obnoxxx:
Locally rebasing this works just fine.
Maybe, just maybe, the auto-merge actually breaks something with the rebasing of layered PRs...

obnoxxx:
@phlogistonjohn - FYI, I reabased locally and updated your PR. no changes.

obnoxxx (approved):
lgtm. 

a few nits in the doc and some thoughts about the comments about the domains, but no reason not to move on for now. 😄

PR #40, opened 2021-03-10
Add first integration tests

phlogistonjohn:
Based on PR #39 please review that first.

These changes add the very first integration tests - tests that expect to be run with a real kubernetes instance and all that fun stuff.
However, these are just the very first integration tests... not a full framework. They expect to be run with a KUBECONFIG env var set and pointed at a k8s cluster that is all set up, including access to an AD instance that matches the hard coded values in the test. The operator must also already be running.

Basically, this is a translation to Go of hand rolled stuff I have been using to test and demo the previous PRs that added the new CRDs. The plan is to build around these tests adding support for bringing up a test env and deploying the operator, etc. This is just a the first steps.

PR #41, opened 2021-03-18
controllers: fix rbac annotation to include config maps

phlogistonjohn:
The operator tries to manage configmap(s) for the samba container
configuration. Without this change, a proper deployment of the operator
will not function because the operator can not access said configmaps.


obnoxxx (approved):
lgmt

(confirmed working on #40)

PR #43, opened 2021-03-18
Kubernetes level CI test including AD

obnoxxx:
based on #42

### DIFF
--- a/tests/files/samba-ad-server-deployment.yml
+++ b/tests/files/samba-ad-server-deployment.yml
@@ -1,0 +1,56 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: samba-ad-server
+  labels:
+    app: samba-ad
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: samba-ad
+  template:
+    metadata:
+      labels:
+        app: samba-ad
+    spec:
+      containers:
+      - name: samba
+        image: quay.io/samba.org/samba-ad-server:latest
+        securityContext:
+          capabilities:
+            add: ["SYS_ADMIN"]
+        ports:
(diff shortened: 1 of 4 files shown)


obnoxxx:
The ci check fails because the the dc can't be found when trying to join in the init container.
Manually putting a section for the domain into the coredns configmap works, but I couldn't find out how to automate it with kubectl patch, since in all my attempts it complains about '\n' characters in the json value... Still playing with it.

@phlogistonjohn

obnoxxx:
@phlogistonjohn  - I figured it out. Now the patching of coredns works (probably it's possible more elegantly), and the joining and ad-member test work!

tests/files/samba-ad-server-deployment.yml
```diff
@@ -0,0 +1,58 @@
...
+    matchLabels:
+      app: samba-ad
+  template:
+    metadata:
+      labels:
+        app: samba-ad
+    spec:
+      containers:
+      - name: samba
+        # Pointing to locally built image,
```
phlogistonjohn, line 19: Comment doesn't seem right to me.

phlogistonjohn (changes requested):
Looks awesome, but I do see one tiny issue I think we should fix as it could be confusing in the future. Thanks!

phlogistonjohn:
> I figured it out. Now the patching of coredns works (probably it's possible more elegantly), and the joining and ad-member test work!

I spent some time on last friday looking for such a more elegant way, and I didn't find one, so this is pretty fine IMO.

tests/files/samba-ad-server-deployment.yml
obnoxxx: correct. fixed :-)

PR #45, opened 2021-03-19
Add liveness checks

phlogistonjohn:
Add liveness checks and update tests to reflect that we now have liveness checks.

Fix a minor annoyance found while testing this fix.


obnoxxx (approved):
lgtm

PR #47, opened 2021-03-23
workflows: run all the tests together via the wrapper script

phlogistonjohn:
Now that we have the ability to run AD in our test setup it doesn't
make a lot of sense to try and individually run each test suite.
Move the script setting up the prerequisites to before the operator
deployment and use our wrapper script.


obnoxxx (approved):
lgtm

PR #48, opened 2021-03-26
Push towards making the integration tests more self sufficent

phlogistonjohn:
These changes work towards making the Go(lang) parts of the test suite able to launch and verify the samba operator.

A lot of the patches simply add infrastructure to allow running the various "phases" of the integration test in order, as well as teaching the tests how to make use of kustomize so that they don't need to actually execute make (ick, :-)).

Unfortunately, these changes do not entirely stand on their own (yet). Certain aspects of the current build process appear to be working due to side effects (or something) and so I can not yet remove the rule that executes make deploy in the CI. Once I understand how to use k3d in the CI more betterer I think we can adapt the CI o rely only on the built artifacts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ ifeq ($(CONTAINER_CMD),)
 	CONTAINER_CMD:=$(shell podman version >/dev/null 2>&1 && echo podman)
 endif
 
-all: manager
+all: manager build-integration-tests
 
 # Run tests
 test: generate vet manifests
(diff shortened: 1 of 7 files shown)


raghavendra-talur:
I have done one pass of the code review and nothing really stands out requiring a change.

I still need to run the tests myself and use the variables to see if there is anything lacking in the UX.

raghavendra-talur (approved):
Tried it once on my laptop. LGTM

PR #49, opened 2021-03-29
Use new 'samba1.sink.test' domain

phlogistonjohn:
To match samba-in-kubernetes/samba-container#40 

Leaving this in DRAFT until the changes linked above are merged and we have a new container image to consume.


tests/files/coredns-snippet.template
```diff
@@ -1,4 +1,4 @@
-    x.asynchrono.us:53 {
+    samba1.sink.test:53 {
```
obnoxxx, line 1: I changed the container repo PR https://github.com/samba-in-kubernetes/samba-container/pull/40 to use `domain1` instead `samba1`. One problem is that the dc's netbios name is `samba1` too, which is problematic.

In general, I prefer using names that tell what kind of object the named object is, e.g. `domain1`, `domain2` etc for domain names `dc1`, `dc2` etc for domain controllers. (I changed the pr to also update the dc's netbios name fwiw.)

obnoxxx (changes requested):

tests/files/smbsecurityconfig2.yaml
```diff
@@ -18,7 +18,7 @@ metadata:
   name: adsec1
 spec:
   mode: active-directory
-  realm: zzz-best.x.asynchrono.us
+  realm: samba1.sink.test
```
obnoxxx, line 21: unrelated to this change, but is it correct to have this in lower case?

phlogistonjohn: Yes, the operator itself is responsible for any case conversions that are needed (if any).

PR #50, opened 2021-03-30
Make integration tests test the operator code in the current branch

phlogistonjohn:
Currently, the CI job for doing integration tests in kubernetes (k3d) only pulls the public complete image from quay.io. This is probably not what we want. :-)

This is my first attempt at getting "locally" built images into the k3d cluster under test and making it part of the integration tests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -82,17 +82,51 @@ jobs:
     #runs-on: ubuntu-latest
     # need to explicitly use 20.04 to avoid problems with jq...
     runs-on: ubuntu-20.04
+    env:
+      CONTAINER_CMD: docker
+      PR_NUM: ${{ github.event.pull_request.number }}
     steps:
     - uses: actions/checkout@v2
-    - uses: nolar/setup-k3d-k3s@v1
+    - name: Install k3d
+      run: curl -L --silent --fail "https://raw.githubusercontent.com/rancher/k3d/main/install.sh" | bash
+    # The TAG env var can interfere with the k3d install script.
+    # Env vars must be set up *after* k3d is installed.
+    - name: Set environment vars
+      run: |
+        tag="latest-scratch"
+        if [ "${PR_NUM}" ] ; then tag="pr-${PR_NUM}" ; fi
+        reg_base="registry.localhost"
+        reg_port="5000"
+        reg="k3d-${reg_base}:${reg_port}"
+        img="${reg}/samba.org/samba-operator:${tag}"
+        {
+          # registry params
+          echo "REG_BASE=${reg_base}"
+          echo "REG_PORT=${reg_port}"
+          echo "REGISTRY=${reg}"
+          # tag and image name for build+push
+          echo "TAG=${tag}"
+          echo "IMG=${img}"
+          # operator image for verification by test suite
+          echo "SMBOP_TEST_EXPECT_MANAGER_IMG=${img}"
+        } >> $GITHUB_ENV
+    - name: Create k3d registry
+      run: k3d registry create "${REG_BASE}" --port "${REG_PORT}"
+    - name: Create k3d cluster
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
> Dang, I missed that.
> 
> In the samba-container repo, I solved this by using the `k3d image import` command:
> https://github.com/samba-in-kubernetes/samba-container/blob/master/.github/workflows/container-image.yml#L58
> and using an `imagePullPolicy` of `Never`: https://github.com/samba-in-kubernetes/samba-container/blob/master/.github/workflows/container-image.yml#L58
> 
> Building a registry may conceptually be better... smile

Well, it almost works but as I re-read the error messages and the doc I used [1] I see that the registry needs to be started *before* the "cluster". But our cluster is being started as part of a 3rd party github action. So we need to somehow deal with this aspect before any of this can work.

obnoxxx:
> > Dang, I missed that.
> > In the samba-container repo, I solved this by using the `k3d image import` command:
> > https://github.com/samba-in-kubernetes/samba-container/blob/master/.github/workflows/container-image.yml#L58
> > and using an `imagePullPolicy` of `Never`: https://github.com/samba-in-kubernetes/samba-container/blob/master/.github/workflows/container-image.yml#L58
> > Building a registry may conceptually be better... smile
> 
> Well, it almost works but as I re-read the error messages and the doc I used [1] I see that the registry needs to be started _before_ the "cluster". But our cluster is being started as part of a 3rd party github action. So we need to somehow deal with this aspect before any of this can work.

What is `[1]` ? 😃 

So `k3d cluster create` can take args and create a registry right when creating the cluster `--registry-create` etc.
Furthermore, the action can pass additional args: https://github.com/nolar/setup-k3d-k3s#k3d-args

so I guess we can use something like this:

```
...
      - uses: nolar/setup-k3d-k3s@v1
        with:
          k3d-args: --registry-create
...
```

phlogistonjohn:
> > > Dang, I missed that.
> > > In the samba-container repo, I solved this by using the `k3d image import` command:
> > > https://github.com/samba-in-kubernetes/samba-container/blob/master/.github/workflows/container-image.yml#L58
> > > and using an `imagePullPolicy` of `Never`: https://github.com/samba-in-kubernetes/samba-container/blob/master/.github/workflows/container-image.yml#L58
> > > Building a registry may conceptually be better... smile
> > 
> > 
> > Well, it almost works but as I re-read the error messages and the doc I used [1] I see that the registry needs to be started _before_ the "cluster". But our cluster is being started as part of a 3rd party github action. So we need to somehow deal with this aspect before any of this can work.
> 
> What is `[1]` ? smiley

The missing link. https://k3d.io/usage/guides/registries/ I meant to link to that so anyone could see what I was basing my changes on.

> 
> So `k3d cluster create` can take args and create a registry right when creating the cluster `--registry-create` etc.
> Furthermore, the action can pass additional args: https://github.com/nolar/setup-k3d-k3s#k3d-args
> 
> so I guess we can use something like this:
> 
> ```
> ...
>       - uses: nolar/setup-k3d-k3s@v1
>         with:
>           k3d-args: --registry-create
> ...
> ```

Yup. That might work and I will try that next. However, I was concerned because I could not figure out what port it was going to use. The registry create command has a -p option I was using. Anyway, I will give it a try soon and we'll see what happens.

phlogistonjohn:
I had to drop the use of the github action for k3d, it was nice but currently has no support for setting up a registry with a predetermined name and port. Not having a predetermined name makes configuring the test very difficult, so I swapped it out.

@obnoxxx , please feel free to review this approach while it is still in draft. Once we've reviewed PR #48 - which this PR depends on, I will take this out of draft.

phlogistonjohn:
Rebased. PTAL @obnoxxx @raghavendra-talur

raghavendra-talur (approved):

PR #51, opened 2021-04-01
push images to an external "scratch" repo

phlogistonjohn:
This is an alternative approach to #50. It is just an experiment currently.


phlogistonjohn:
Idea does not work. Secrets can not be used from forks, the normal contribution workflow. This makes this approach fairly useless. Closing.

PR #52, opened 2021-04-08
Create services for access to servers

phlogistonjohn:
First, the patches define & name a Server Group concept, implied as part of the Phase 1 design doc. This is the name of smb server (or group of servers) hosting the share defined by a SmbShare resource.
Then this name is used to properly label and create kubernetes Services to expose port 445 for the instance.
A new test case, using the k8s internal dns name for the service is added.


phlogistonjohn:
NOTE: This PR is a draft because it requires the work in PR #50 - as it currently stands the new tests are tested against the "latest" tag of the quay.io repo - not the changes in this PR!
Until that PR is merged this PR will always fail.

internal/resources/smbshare.go
```diff
@@ -148,6 +148,15 @@ func (m *SmbShareManager) Update(ctx context.Context, instance *sambaoperatorv1a
 		return Requeue
 	}
 
+	_, created, err = m.getOrCreateService(
+		ctx, planner, instance.Namespace)
+	if err != nil {
+		return Result{err: err}
+	} else if resized {
```
raghavendra-talur, line 155: s/resized/created

raghavendra-talur (changes requested):
Except for one small change, everything else looks very good!

internal/resources/smbshare.go
phlogistonjohn: fixed

phlogistonjohn:
@raghavendra-talur or @obnoxxx  PTAL. We'll need this as a starting point for future work.

Issue #56, opened 2021-04-30
samba-operator-controller-manager continiously restarting with status OOMKilled 


nixpanic:
After deploying the operator with `make deploy` on OpenShift, the pod is restarting constantly:

```
$ kubectl -n samba-operator-system get pods
NAME                                                READY   STATUS      RESTARTS   AGE
samba-operator-controller-manager-5c4766cfc-hnp2k   1/2     OOMKilled   2          5m58s
$ kubectl -n samba-operator-system describe pods
Name:         samba-operator-controller-manager-5c4766cfc-hnp2k
Namespace:    samba-operator-system
Priority:     0
Node:         ip-10-0-153-215.ec2.internal/10.0.153.215
Start Time:   Fri, 30 Apr 2021 10:18:13 +0200
Labels:       control-plane=controller-manager
              pod-template-hash=5c4766cfc
Annotations:  k8s.v1.cni.cncf.io/network-status:
                [{
                    "name": "",
                    "interface": "eth0",
                    "ips": [
                        "10.131.0.62"
                    ],
                    "default": true,
                    "dns": {}
                }]
              k8s.v1.cni.cncf.io/networks-status:
                [{
                    "name": "",
                    "interface": "eth0",
                    "ips": [
                        "10.131.0.62"
                    ],
                    "default": true,
                    "dns": {}
                }]
              openshift.io/scc: restricted
Status:       Running
IP:           10.131.0.62
IPs:
  IP:           10.131.0.62
Controlled By:  ReplicaSet/samba-operator-controller-manager-5c4766cfc
Containers:
  kube-rbac-proxy:
    Container ID:  cri-o://9dcea85dea85669f5d32af871cc81f45dd0ab2a98f762d8d3763afd3d9c5d5f2
    Image:         gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0
    Image ID:      gcr.io/kubebuilder/kube-rbac-proxy@sha256:e10d1d982dd653db74ca87a1d1ad017bc5ef1aeb651bdea089debf16485b080b
    Port:          8443/TCP
    Host Port:     0/TCP
    Args:
      --secure-listen-address=0.0.0.0:8443
      --upstream=http://127.0.0.1:8080/
      --logtostderr=true
      --v=10
    State:          Running
      Started:      Fri, 30 Apr 2021 10:18:18 +0200
    Ready:          True
    Restart Count:  0
    Environment:    <none>
    Mounts:
[...]

PR #57, opened 2021-04-30
deploy: increase memory limit to 100Mi for the manager

nixpanic:
When running on OpenShift (where the resource limits are enforced by
default), the manager container gets OOM killed. It seems that the limit
of 30Mi is too low, when starting the container consumes close to 70Mi
after which it reduces back to 40Mi.

Fixes: #56


phlogistonjohn (approved):
Change looks good, thanks.

I am a little bit concerned that the operator is so young but already hits the resource limit the sdk defaults to in the template.

Issue #11, opened 2020-10-23
Smbd container as configured by operator fails in openshift


phlogistonjohn:
OpenShift's extra security stuff seems to cause the samba container to fail. We'll need to properly research this in order to either code up or write up how to run on openshift.

nixpanic:
I might be hitting an issue related to OpenShift restrictions too.
After running `make deploy` and creating a new SmbShare from the `tests/` directory:

```
2021-04-29T14:31:26.345Z	ERROR	controllers.SmbShare	Failed to create new Deployment	{"smbshare": "sink/tshare1", "Deployment.Namespace": "sink", "Deployment.Name": "tshare1", "error": "deployments.apps \"tshare1\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}
github.com/go-logr/zapr.(*zapLogger).Error
	/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128
github.com/samba-in-kubernetes/samba-operator/internal/resources.(*SmbShareManager).getOrCreateDeployment
	/workspace/internal/resources/smbshare.go:193
github.com/samba-in-kubernetes/samba-operator/internal/resources.(*SmbShareManager).Update
	/workspace/internal/resources/smbshare.go:123
github.com/samba-in-kubernetes/samba-operator/internal/resources.(*SmbShareManager).Process
	/workspace/internal/resources/smbshare.go:78
github.com/samba-in-kubernetes/samba-operator/controllers.(*SmbShareReconciler).Reconcile
	/workspace/controllers/smbshare_controller.go:54
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.2/pkg/internal/controller/controller.go:235
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.2/pkg/internal/controller/controller.go:209
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.2/pkg/internal/controller/controller.go:188
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1
[...]

PR #58, opened 2021-04-30
deploy/rbac: allow the controller to set smbshares/finalizers

nixpanic:
Creating an SmbShare on OpenShift fails with

    Failed to create new PVC	{"smbshare": "samba-operator-system/tshare1", "pvc.Namespace": "samba-operator-system", "pvc.Name": "tshare1-pvc", "error": "persistentvolumeclaims \"tshare1-pvc\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}

This is caused by the OwnerReferencesPermissionEnforcement plugin that
is enabled by default on OpenShift.

By adding the permissions for the controller to set finalizers on the
SmbShare resource, this problem does not occur anymore.

Updates: #11

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -84,6 +84,14 @@ rules:
 - apiGroups:
   - samba-operator.samba.org
   resources:
+  - smbshares/finalizers
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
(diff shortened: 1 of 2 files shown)


phlogistonjohn (approved):
Thanks a ton for looking into this!

PR #60, opened 2021-05-06
design: Add DNS related subsection to SmbSecurityConfig

phlogistonjohn:
Add a subsection to the SmbSecurityConfig CR design so that
we can tell the operator we want to configure Pods such that
they register with the AD DNS. Currently, only two behaviors
are supported:
* never - do not ever register with AD DNS (default)
* external-ip - discover the Pod's external ip [1] and register
  it with AD DNS

[1] - This would currently require a Service with type LoadBalancer.


raghavendra-talur (approved):

phlogistonjohn:
Test failure appears spurious and this is a doc-only change. Merging.

PR #65, opened 2021-05-12
dockerfile: use an up-to-date golang builder image

phlogistonjohn:
The build image was created in the olden days and has not been updated
in a while. So it is not using a currently supported version of golang.
This updates it to the latest golang version.


gd (approved):
LGTM, thanks!

PR #70, opened 2021-05-19
Initial support for events

phlogistonjohn:
Add initial support for recording events on the SmbShare resource as the operator reconciles things and creates resources to support the share.


config/rbac/role.yaml
```diff
@@ -6,6 +6,13 @@ metadata:
   creationTimestamp: null
   name: manager-role
 rules:
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - create
+  - patch
```
nixpanic, line 15: Why is `create` not sufficient?

phlogistonjohn: I dunno. I copied it from the kubebuilder book: https://book-v1.book.kubebuilder.io/beyond_basics/creating_events.html

controllers/smbshare_controller.go
```diff
@@ -45,14 +47,16 @@ type SmbShareReconciler struct {
 // +kubebuilder:rbac:groups=core,resources=persistentvolumeclaims,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
```
raghavendra-talur, line 50: I think the groups value should be core here too.

raghavendra-talur (changes requested):

controllers/smbshare_controller.go
phlogistonjohn: OK, will do

phlogistonjohn:
I have removed the "patch" verb, made it use core rather than "", and added a test case where there was none before. PTAL.

raghavendra-talur (approved):

PR #77, opened 2021-06-12
Support for AD DNS registration

phlogistonjohn:
Initial support for registering a pods (public/external facing) IP address with the AD DNS.
This involves enabling the feature within the SmbSecurity Config. Then the operator will add two new containers to the pod, one to monitor the Service for it's IP address and one to submit requests to AD to register said IP(s) with AD.
The pod does all of the work, the major reason being is that only the (joined) pod can make authenticated requests to the Active Directory to make the changes.

A simplistic test is also added.


api/v1alpha1/smbsecurityconfig_types.go
```diff
@@ -97,6 +102,16 @@ type SmbSecurityDomainSpec struct {
 	// TODO: add support for id mapping ranges, etc.
 }
 
+// SmbSecurityDNSSpec configures the relationship between systems managed
+// vi this SmbSecurityConfig and the domain. Ignored by user mode.
```
raghavendra-talur, line 106: s/vi/via

raghavendra-talur (approved):

api/v1alpha1/smbsecurityconfig_types.go
phlogistonjohn: good catch thanks!

PR #78, opened 2021-06-15
docs: update the devloper on how to set env vars

phlogistonjohn:
Now that the base deployment yaml has a pre-existing env key, the
previous example no longer applies correctly. It overwrites the
needed namespace var. Update the patch to append to the list
at the env key.

### DIFF
--- a/docs/developers-notes.md
+++ b/docs/developers-notes.md
@@ -65,19 +65,22 @@ We will set the environment variables using kustomize rules in the file
 patches:
 - patch: |-
     - op: add
-      path: /spec/template/spec/containers/0/env
+      path: /spec/template/spec/containers/0/env/-
       value:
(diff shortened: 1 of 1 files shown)


docs/developers-notes.md
```diff
@@ -65,19 +65,22 @@ We will set the environment variables using kustomize rules in the file
 patches:
 - patch: |-
     - op: add
-      path: /spec/template/spec/containers/0/env
+      path: /spec/template/spec/containers/0/env/-
       value:
-      - name: "SAMBA_OP_SMBD_CONTAINER_IMAGE"
```
raghavendra-talur, line 70: Is this removal of '-' intentional?

raghavendra-talur: Never mind, I found the answer in the link you provided below. It looks good to me.

raghavendra-talur (approved):

Issue #87, opened 2021-08-03
Creating SMBShares in different namespaces


ygalblum:
Looking at the code, the operator creates the resources (PVC, Deployment, Service) in the operator's namespace and then sets the SMBShare resource as the owner of these resources (to have K8S garbage collect them).  

But, K8S does not allow cross namespace ownership. So, if the SMBShare is created in a different namespace, this operation will fail. Please note that the return value is not checked: https://github.com/samba-in-kubernetes/samba-operator/blob/878ac25673c0fd7d3a2f2294f8714843c6179c86/internal/resources/smbshare.go#L311

In the demo showed in SambaXP right after provisioning the operator the context is set to use the namespace of the operator. As a result, all the resources created in the demo are created in that namespace and this issue does not occur.

Is the intention to force the users to create SMBShares only in the operator's namespace?

phlogistonjohn:
Hi!  First off, thanks a lot for this feedback it helps a lot.

> Looking at the code, the operator creates the resources (PVC, Deployment, Service) in the operator's namespace and then sets the SMBShare resource as the owner of these resources (to have K8S garbage collect them).
> 

To be honest, this code has gone back and forth a bit and is quite immature. I am not really shocked to find we're not handling this right.

> But, K8S does not allow cross namespace ownership. So, if the SMBShare is created in a different namespace, this operation will fail. Please note that the return value is not checked:
> 
> https://github.com/samba-in-kubernetes/samba-operator/blob/878ac25673c0fd7d3a2f2294f8714843c6179c86/internal/resources/smbshare.go#L311
> 

This is very helpful thanks - it reveals two problems: one is what you point out and the other is that our static checker is not detecting ignored errors!


> In the demo showed in SambaXP right after provisioning the operator the context is set to use the namespace of the operator. As a result, all the resources created in the demo are created in that namespace and this issue does not occur.
> 
> Is the intention to force the users to create SMBShares only in the operator's namespace?

If you want to try out the samba share's operation it is true that creating the resources in the same namespace as the operator itself is your best bet today. In most of our integration tests this is what's done and it generally works.

What we ultimately want to do is a bit "squishy" as we're still trying to learn how best to deploy the various samba components in k8s and I'm not sure what workflows we want to support. That's where feedback like yours is valuable. If you were deploying this operator in your cluster, how would you like it to work in the ideal case?


[...]

PR #88, opened 2021-08-03
resources: fix failure to check returned errors from SetControllerReference

phlogistonjohn:
This also makes the error logging calls consistent between the calls to
SetControllerReference and the API calls to create/update/whatever the
resource. This eliminates the functions that existed to create a
resource type and then set the controller reference as it was easier
to inline these "two line" functions than plumb error handling through
everything.

NOTE: I expect this may cause the CI to fail. If so, that's a good thing. :-)
But I want to see it first.

### DIFF
--- a/internal/resources/smbshare.go
+++ b/internal/resources/smbshare.go
@@ -232,7 +232,21 @@ func (m *SmbShareManager) getOrCreateDeployment(
 
 	if errors.IsNotFound(err) {
 		// not found - define a new deployment
-		dep := m.deploymentForSmbShare(planner, ns)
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
Yup it failed. This means that issue #87 is on the right track and my recent changes in this area were incorrect. The good thing is that the CI actually catches a problem (tests were added as part of that work) once we add the appropriate error checking to the code.

Now's just a matter of deciding what to do for long term behavior wrt namespaces.

phlogistonjohn:
Obsoleted by #100. Closing.

PR #92, opened 2021-09-01
docs: general documentation update

phlogistonjohn:
The readme and our minimal docs were a bit out of date. This fleshes out
the structure of our docs while updating them for the present day.

### DIFF
--- a/docs/howto.md
+++ b/docs/howto.md
@@ -1,0 +1,293 @@
+
+# Create a simple share
+
+The simplest thing one can do with the operator is to create a single SmbShare
+with an embedded PVC spec:
+
+```yaml
+apiVersion: samba-operator.samba.org/v1alpha1
+kind: SmbShare
+metadata:
+  name: myshare
+spec:
+  readOnly: false
+  storage:
+    pvc:
+      spec:
+        accessModes:
+          - ReadWriteOnce
+        resources:
+          requests:
+            storage: 1Gi
+```
+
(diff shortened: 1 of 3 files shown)


phlogistonjohn:
@rravi6121 since you've been digging into the operator and trying things out, I'd love your feedback here as well.

docs/howto.md
```diff
@@ -0,0 +1,292 @@
...
+    publish: external
+```
+```yaml
+apiVersion: samba-operator.samba.org/v1alpha1
+kind: SmbShare
+metadata:
+  name: myshare
+spec:
+  securityConfig: mydomain
+  readOnly: false
```
rravi6121, line 279: @phlogistonjohn commonconfig needs to be included here

phlogistonjohn: Fixed. Thanks for catching that!

docs/howto.md
```diff
@@ -0,0 +1,293 @@
...
+      spec:
+        accessModes:
+          - ReadWriteOnce
+        resources:
+          requests:
+            storage: 1Gi
+```
+
+Now the name of the SMB share will be "My Great Share" instead of "myshare".
+Note that this only effects the name of the share in the SMB protocol.
```
raghavendra-talur, line 53: more apt to use affects instead of effects here?

raghavendra-talur (approved):

PR #94, opened 2021-09-16
tests: use k8s api for executing smbclient commands

### LINKED ISSUE #93: kubectlSmbClientCli in the test utils uses kubeconfig ENV as --kubeconfig value
Sometimes, it is not right to do that, especially in the case when the kubeconfig ENV is of the form file1:file2:file3 etc.

I think the solution in this case is probably to skip providing the arg to kubectl *if* the ENV is set.

phlogistonjohn:
Previously, we were shelling out to `kubectl exec` to run smbclient commands on our smbclient test pod. These changes add a new api based TestExec type to the `test/utils/kube` package. Then it replaces the kubectl based version with a new version based on TestExec and its types.

This should be more effieicent as well as hopefully fix #93 because the old implementation was (ab)using KUBECONFIG in a way that broke on some setups.

fixes: #93


phlogistonjohn:
Before we fix the other reason the ci is failing:
@raghavendra-talur (and anyone else interested) please take a look at these failing tests as it reveals the new errors that you get based on this new code. It includes more information than before as well as quoting the command in a semi-useful manner.

raghavendra-talur (dismissed):

raghavendra-talur (approved):

PR #97, opened 2021-09-23
Use latest controller-gen code to avoid build problems

gd:
Guenther

Signed-off-by: Günther Deschner <gd@samba.org>


phlogistonjohn:
I tried this yesterday and it was not working for me (not submitted as a PR yet, for various reasons). The CI errors jives with what I saw... this latest version also changes the CLI options. You can try either going 1 version back which worked for me, or try fixing the CLI -- but I did the former because I couldn't find any docs explaining the new cli (and the built in help for controller-gen is rather hard to understand IMO).

phlogistonjohn:
@gd how do you prefer to move forward with this one?

gd:
Given the clarification we had later on I think we can close this one.

PR #99, opened 2021-09-23
workflows: force a version of k3s that should fix our ci issues

phlogistonjohn:
See: https://github.com/k3s-io/k3s/issues/3704

Figuring out what to actually put on the --image option was far harder
than it seems it should have been. :-\

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -113,7 +113,7 @@ jobs:
     - name: Create k3d registry
       run: k3d registry create "${REG_BASE}" --port "${REG_PORT}"
     - name: Create k3d cluster
-      run: k3d cluster create --wait --registry-use "${REGISTRY}"
+      run: k3d cluster create --wait --image docker.io/rancher/k3s:v1.21.5-k3s1 --registry-use "${REGISTRY}"
     - name: Wait for cluster ready
       run: |
         while ! kubectl get serviceaccount default >/dev/null; do sleep 1; done


.github/workflows/main.yml
```diff
@@ -113,7 +113,7 @@ jobs:
     - name: Create k3d registry
       run: k3d registry create "${REG_BASE}" --port "${REG_PORT}"
     - name: Create k3d cluster
-      run: k3d cluster create --wait --registry-use "${REGISTRY}"
+      run: k3d cluster create --wait --image docker.io/rancher/k3s:v1.21.5-k3s1 --registry-use "${REGISTRY}"
```
raghavendra-talur, line 116: Instead of hardcoding to one particular patch version, would it make more sense to lock on to a Minor version?

Something like
```
IMAGE_TAG=$(curl --silent "https://update.k3s.io/v1-release/channels/v1.21" | egrep -o '/v[^ ]+"' | sed -E 's/\/|\"//g' | sed -E 's/\+/\-/')
k3d cluster create --wait --image docker.io/rancher/k3s:${IMAGE_TAG} --registry-use "${REGISTRY}"
```

raghavendra-talur (approved):

.github/workflows/main.yml
phlogistonjohn: Sure, let's follow up with something like that in a follow-up pr.

PR #100, opened 2021-09-24
Fix namespace behavior

### LINKED ISSUE #87: Creating SMBShares in different namespaces

phlogistonjohn:
Fixes #87

These changes reverse the well intentioned but incorrect attempt at managing all created resources under the namespace the operator runs in. It restores the previous behavior such that deployments/pods/services/etc will be created in the namespace that the SmbShare resource "owning" those resources is placed in.

This also fixes up the existing "other namespace" test to conform to the new behavior.

### DIFF
--- a/internal/resources/smbshare.go
+++ b/internal/resources/smbshare.go
@@ -117,7 +117,7 @@ func (m *SmbShareManager) Update(
 		return Requeue
 	}
 
-	destNamespace := m.cfg.WorkingNamespace
+	destNamespace := instance.Namespace
 	cm, created, err := getOrCreateConfigMap(ctx, m.client, destNamespace)
 	if err != nil {
 		return Result{err: err}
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
found a bug by re reading my code. moving to draft.

internal/resources/smbshare.go
```diff
@@ -191,7 +191,8 @@ func (m *SmbShareManager) Finalize(
 	ctx context.Context,
 	instance *sambaoperatorv1alpha1.SmbShare) Result {
 	// ---
-	cm, err := getConfigMap(ctx, m.client, m.cfg.WorkingNamespace)
+	destNamespace := instance.Namespace
```
raghavendra-talur, line 194: IMO, it would be easier to read if you used instance.Namespace as an argument in the next line.

raghavendra-talur (approved):

internal/resources/smbshare.go
phlogistonjohn: It's there to help me remember things better. I'll think I will leave it for now, thanks.

raghavendra-talur (approved):

PR #101, opened 2021-09-24
Update some dependencies & related tweaks

phlogistonjohn:
This is a bit of a grab bag, but mainly I'm refreshing the dependencies of libs and tools and generally keeping the build env fresh and consistent. :-)


phlogistonjohn:
@gd this PR overlaps with your pr #97  -  I figure we both did the work around the same time but I wanted to fix the CI before pushing this work. Since your pr uses the latest version but this makes the more modest change to v0.6.2 I figure we can take this and then use your version to figure out what's needed to use the newest controller gen. How's that sound?

raghavendra-talur (approved):

PR #103, opened 2021-09-30
Use multiple config maps for container config

phlogistonjohn:
Depends on #100  - please review that first

The samba-container is based on a JSON configuration file which gets turned into smb/samba configs and environmental config. Originally, we had "one big config" but this was running counter to the flow of the rest of the operator. On top of that the work to fix the namespaces show that we're going to be creating additional config maps when the resources occur in different namespaces anyway. Using a config map per-SmbShare (instance) let's us tie the lifecycle of the map to the SmbShare and simplify the management of the config.

### DIFF
--- a/internal/resources/smbshare.go
+++ b/internal/resources/smbshare.go
@@ -326,6 +326,57 @@ func (m *SmbShareManager) getOrCreatePvc(
 	return nil, false, err
 }
 
+func (m *SmbShareManager) getOrCreateService(
+	ctx context.Context, planner *sharePlanner, ns string) (
+	*corev1.Service, bool, error) {
+	// Check if the service already exists, if not create a new one
+	found := &corev1.Service{}
+	err := m.client.Get(
+		ctx,
+		types.NamespacedName{
+			Name:      planner.instanceName(),
+			Namespace: ns,
+		},
+		found)
+	if err == nil {
+		return found, false, nil
+	}
+
+	if errors.IsNotFound(err) {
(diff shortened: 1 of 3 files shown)


internal/resources/smbshare.go
```diff
@@ -377,6 +377,81 @@ func (m *SmbShareManager) getOrCreateService(
...
+		Name:      planner.instanceName(),
+		Namespace: ns,
+	}
+	err := m.client.Get(ctx, cmKey, found)
+	if err == nil {
+		return found, false, nil
+	}
+
+	if !errors.IsNotFound(err) {
+		// unexpected errror!
```
raghavendra-talur, line 445: Typo for future fix

internal/resources/smbshare.go
```diff
@@ -377,6 +377,81 @@ func (m *SmbShareManager) getOrCreateService(
...
+			err,
+			"Failed to generate default ConfigMap",
+			"SmbShare.Namespace", planner.SmbShare.Namespace,
+			"SmbShare.Name", planner.SmbShare.Name,
+			"ConfigMap.Namespace", cm.Namespace,
+			"ConfigMap.Name", cm.Name)
+		return cm, false, err
+	}
+	// set the smbshare instance as the owner and controller
+	err = controllerutil.SetControllerReference(
```
raghavendra-talur, line 468: I am not sure whether SetControllerReference() is the right function to use here or SetOwnerReference(). I don't know for sure though.

PR #124, opened 2021-11-04
limit switching availabilty modes

### LINKED ISSUE #119: [ctdb/ss] Expliclity handle cases of SmbShare contents changing
In the future we will probably want to fully reconcile certain changes to the SmbShare, such as changing from a non-clustered to a clustered instance. However in the short term it should be enough to recognize the situation and refuse to do anything (too) destructive.

phlogistonjohn:
Fixes #119

Prevent certain, possibly destructive, changes occurring if the availabilityMode field of the scaling spec is changed.

### DIFF
--- a/internal/resources/smbshare.go
+++ b/internal/resources/smbshare.go
@@ -162,7 +162,8 @@ func (m *SmbShareManager) Update(
 		instance.Spec.Storage.Pvc.Name = pvc.Name
 	}
 
-	if instance.Annotations[serverBackend] == "" {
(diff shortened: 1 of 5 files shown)


raghavendra-talur (approved):
The implementation itself looks good to me.

There is a larger question about where to store the previous state in. John and I discussed in a call about what to use annotations vs status for, if the serverBackend setting should be stored in annotation or status.

We have currently chosen to go with the current implementation as this code is gong to change soon again.

PR #129, opened 2021-12-20
Adding vfs_fileid changes to the samba-operator

spuiuk:
Tested by building the operator image, pushing it to a remote repo and then deploying that remote repo. Confirmed the change by checking the operator logs for the commit id. Once the smbshare was created, I did a kubectl exec into it and checked the generated smb.conf file using the net conf list command 

This set of patches contains two changes.
1) Remove mention of No Printing and
2) Add the fileid plugin based on a global conf set. This can be replaced by intelligence to check for vfs_fileid requirements in the future.

### DIFF
--- a/internal/smbcc/container_config.go
+++ b/internal/smbcc/container_config.go
@@ -27,6 +27,12 @@ const (
 	CTDB FeatureFlag = "ctdb"
 )
 
+// GlobalOptions is used to pass options to modify the samba configuration
+type GlobalOptions struct {
+	// AddVFSFileid is used to check if we add vfs_fileid to the smb config
+	AddVFSFileid bool
+}
+
 // SambaContainerConfig holds one or more configuration for samba
 // containers.
 type SambaContainerConfig struct {
@@ -101,6 +107,13 @@ const (
 	No = "no"
 )
 
+// NewGlobalOptions is the constructor for struct SambaConfigOptions
+func NewGlobalOptions() GlobalOptions {
+	return GlobalOptions{
+		AddVFSFileid: true,
+	}
+}
+
 // New returns a new samba container config.
 func New() *SambaContainerConfig {
 	return &SambaContainerConfig{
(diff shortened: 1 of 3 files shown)


internal/smbcc/container_config.go
```diff
@@ -99,6 +99,9 @@ const (
 	Yes = "yes"
 	// No means no.
 	No = "no"
+
+	// AddVFSFileid is used to chec if we add vfs_fileid to the config
+	AddVFSFileid = true
```
phlogistonjohn, line 104: This doesn't seem like the right approach to me, why is this a global rather than say, an argument to NewGlobals?

phlogistonjohn (changes requested):

spuiuk:
Switched to Draft PR.

internal/smbcc/container_config.go
spuiuk: I've added a new commit which changes the way we pass the configuration options. Let me know if this is what you had in mind. Not squashing patches until they are acceptable.

internal/smbcc/container_config.go
```diff
@@ -99,11 +104,21 @@ const (
 	Yes = "yes"
 	// No means no.
 	No = "no"
-
-	// AddVFSFileid is used to chec if we add vfs_fileid to the config
-	AddVFSFileid = true
 )
 
+// Constructor for the OperatorOptions struct
+func NewOperatorOptions() *OperatorOptions {
```
phlogistonjohn, line 110: This is more of a taste thing I think but I'd drop this function and simply not provide a "constructor". I'd also make it a pass-by-value (not a pointer). If you want me to elaborate on why please ask, but I'm lazy (and still on vacation ;-)) so I'm not going to elaborate yet.  However, it's minor and I could easily live with it as-is.

spuiuk: Hah. Sorry, I didn't realise that you would be looking at patches today. I have just posted another one. 
Please excuse my changes as I navigate go lang. Let's schedule a 1-1 tomorrow to go through the code.

spuiuk:
John,
I've made most of the changes we discussed earlier except for the use of a boolean variable in the GlobalOptions. This is a simple check and adding a slice or enum unnecessarily complicates the code and increases the chances of regression. So using the YANNI principal, I'll stick to boolean for now until we have a requirement for additional modules.

phlogistonjohn (commented):
On first glance it looks OK, but I'll give it a more thorough reading during the work week.
Also @raghavendra-talur PTAL.

phlogistonjohn (approved):
LGTM, thanks.

PR #130, opened 2022-01-11
Golangci lint

synarete:
Enable 'golanci-lint' tool via 'make lint' rule.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -20,6 +20,7 @@ linters:
     - structcheck
     - typecheck
     - unconvert
+    - unused
     - varcheck
     - whitespace
 
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -11,6 +11,7 @@ linters:
     - errcheck
     - exhaustive
     - exportloopref
+    - funlen
     - goconst
     - goheader
     - gofmt
@@ -25,4 +26,7 @@ linters:
     - unused
     - varcheck
     - whitespace
-
+linters-settings:
+  funlen:
+    lines: 188
+    statements: 60
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,24 @@
+---
+run:
+  timeout: 3m
+linters:
+  disable-all: true
+  enable:
+    - asciicheck
+    - deadcode
+    - depguard
+    - dupl
+    - exportloopref
+    - goconst
+    - goheader
+    - gofmt
+    - gosimple
+    - govet
+    - megacheck
+    - misspell
+    - structcheck
+    - typecheck
+    - unconvert
+    - varcheck
+    - whitespace
+
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -8,6 +8,7 @@ linters:
     - deadcode
     - depguard
     - dupl
+    - errcheck
     - exportloopref
     - goconst
     - goheader
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -9,6 +9,7 @@ linters:
     - depguard
     - dupl
     - errcheck
+    - exhaustive
     - exportloopref
     - goconst
     - goheader
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -12,20 +12,30 @@ linters:
     - exhaustive
     - exportloopref
     - funlen
+    # - gci
     - goconst
     - gocritic
+    # - godot
+    # - godox
     - goheader
     - gofmt
     - gosimple
     - govet
+    # - ineffassign
+    # - ifshort
     - lll
     - megacheck
     - misspell
+    # - nlreturn
+    # - nolintlint
     - revive
     - staticcheck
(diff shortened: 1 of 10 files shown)


internal/resources/planner.go
```diff
@@ -301,7 +301,7 @@ func (sp *sharePlanner) update() (changed bool, err error) {
 	return
 }
 
-func (sp *sharePlanner) prune() (changed bool, err error) {
+func (sp *sharePlanner) Prune() (changed bool, err error) {
```
phlogistonjohn, line 304: I think we can probably remove this function as it probably is really unused now.

.golangci.yaml
```diff
@@ -9,6 +9,7 @@ linters:
     - depguard
     - dupl
     - errcheck
+    - exhaustive
```
phlogistonjohn, line 12: I'm a tad bit ambivalent about this value of this one but I won't object.

.golangci.yaml
```diff
@@ -25,4 +26,7 @@ linters:
     - unused
     - varcheck
     - whitespace
-
+linters-settings:
+  funlen:
+    lines: 96
```
phlogistonjohn, line 44: Where do these values come from?
Would it be better to adjust these to match the current code rather than mark some of the existing functions not to be checked?

.golangci.yaml
```diff
@@ -18,6 +18,7 @@ linters:
     - gofmt
     - gosimple
     - govet
+    - lll
```
phlogistonjohn, line 26: What a terrible name. :-)

.golangci.yaml
```diff
@@ -31,3 +32,5 @@ linters-settings:
   funlen:
     lines: 96
     statements: 48
+  lll:
+    line-length: 100
```
phlogistonjohn, line 47: This disagrees with the line length set in the revive config.  On top of that, is it not redundant if revive is already checking line length?

.golangci.yaml
```diff
@@ -34,3 +35,7 @@ linters-settings:
     statements: 48
   lll:
     line-length: 100
+  revive:
+    rules:
+      - name: var-naming
+        arguments: [["UID", "GID"]]
```
phlogistonjohn, line 51: Can this consume our existing revive config file rather than reproducing the settings here?

phlogistonjohn (commented):
Looking good in general. I have some specific questions/comments inline.

phlogistonjohn:
One additional thought... should we have one top level `make check` that runs all static check tools (run individual tools like `make check-revive` or `make check-foo`)?

internal/resources/planner.go
synarete: Would this function be needed in the future? If so, maybe it is better to leave it as is and just add "// nolint:unused" .

.golangci.yaml
synarete: Same set of arguments in-favor/against using gcc '-Wswitch-enum' in C/C++ (which indeed, is not enabled automatically by '-Wall')

.golangci.yaml
synarete: OK, will align those values with current code (lines: 188, statements: 60)

.golangci.yaml
synarete: Fixed to lll.line-length=98 (same as revive). Apparently, this linter overlaps checks done by revive.

.golangci.yaml
synarete: Apparently, there isn't such option. I merged the rest of revive's toml config-options into golangci.yaml

synarete:
> One additional thought... should we have one top level `make check` that runs all static check tools (run individual tools like `make check-revive` or `make check-foo`)?

Agreed. Fixed Makefile accordingly (i.e.: `make check` calls all static check tools).

.golangci.yaml
phlogistonjohn: OK, that's a bit disappointing to hear. I won't object to it in this PR but I will reserve the right to do something about the duplication in the future if keeping these configs in sync becomes burdensome.

Hopefully before that happens someone (us?) could contribute to the linter the ability to pass a path :-)

internal/resources/planner.go
phlogistonjohn: I'm fairly certain that it's now unused as I abandoned the earlier appraoch of trying to use a shared config map across multiple instances. However, don't feel obligated to clean that up now as long as the function is left private, I'd be fine with adding the special comment. We can double check that it's unused and remove it in a future pr.

.golangci.yaml
phlogistonjohn: Awesome, thanks!

phlogistonjohn (approved):
While I have some quibbles with some of the checks (gocritic in particular) I don't think its anything that should block the PR. Thanks!

phlogistonjohn (approved):

raghavendra-talur (approved):

PR #131, opened 2022-01-16
Yamllint

synarete:
Require all yaml files to be well formatted, using 'yamllint' tool. If need, reformat yaml file using 'yq'.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -231,3 +231,17 @@ GOLANGCI_LINT=$(GOBIN)/golangci-lint
 else
 GOLANGCI_LINT=$(shell command -v golangci-lint ;)
 endif
+
+# find or download yq
+.PHONY: yq
+yq:
+ifeq (, $(shell command -v yq ;))
+	@echo "yq not found in PATH, checking in GOBIN ($(GOBIN))..."
+ifeq (, $(shell command -v $(GOBIN)/yq ;))
+	$(GO_CMD) install github.com/mikefarah/yq/v4@latest
+	@echo "yq installed in GOBIN"
+endif
+YQ=$(GOBIN)/yq
+else
+YQ=$(shell command -v yq ;)
+endif
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,7 @@ GO_CMD:=go
 GOFMT_CMD:=gofmt
 KUBECTL_CMD:=kubectl
 BUILDAH_CMD:=buildah
+YAMLLINT_CMD:=yamllint
 
 # Image URL to use all building/pushing image targets
 TAG ?= latest
@@ -121,6 +122,11 @@ reformat:
 vet:
 	$(GO_CMD) vet ./...
 
+# Format yaml files for yamllint standard
+.PHONY: yaml-fmt
+yaml-fmt: yq
+	YQ=$(YQ) ./hack/yq-fixup-yamls.sh ./
+
 # Generate code
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
@@ -186,9 +192,9 @@ bundle-build:
 	@false
 	# See vcs history for how this could be restored or adapted in the future.
 
-.PHONY: check check-revive check-golangci-lint check-format
+.PHONY: check check-revive check-golangci-lint check-format check-yaml
 
-check: check-revive check-golangci-lint check-format vet
+check: check-revive check-golangci-lint check-format vet check-yaml
 
 check-format:
 	! $(GOFMT_CMD) $(CHECK_GOFMT_FLAGS) . | sed 's,^,formatting error: ,' | grep 'go$$'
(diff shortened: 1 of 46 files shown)


hack/yq-fixup-yamls.sh
```diff
@@ -0,0 +1,40 @@
...
+	dif=$(diff -q "${yaml}" "${yaml}".tmp)
+	if [ "${dif}" != "" ]; then
+		mv "${yaml}".tmp "${yaml}"
+	else
+		rm -f "${yaml}".tmp
+	fi
+}
+
+_yaml_fixup_files() {
+	yamls=$(find "$1" -type f -name '*.yml' -or -name '*.yaml')
```
phlogistonjohn, line 27: Perhaps we should do -iname instead of name for case-insensitive matches? (example: FOO.YAML)

Makefile
```diff
@@ -28,6 +28,7 @@ GO_CMD:=go
 GOFMT_CMD:=gofmt
 KUBECTL_CMD:=kubectl
 BUILDAH_CMD:=buildah
+YAMLLINT_CMD:=yamllint
```
phlogistonjohn, line 31: Unlike YQ there's no rule to download yamllint if it is not found. Why's that?

phlogistonjohn (changes requested):
Generally, looking good. I didn't dig deeply into the yamllint rules or the reformatted YAMLs and just assumed the tools were doing the right thing.

hack/yq-fixup-yamls.sh
```diff
@@ -0,0 +1,40 @@
...
+#   yq-fixup-yamls.sh dirpath
+#
+YQ=${YQ:-$(command -v yq)}
+${YQ} --version > /dev/null
+
+_yaml_fixup_file() {
+	yaml="$1"
+
+	# use yq as yaml formatter
+	${YQ} eval "${yaml}" > "${yaml}".tmp
```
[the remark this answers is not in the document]
synarete: ACK

hack/yq-fixup-yamls.sh
```diff
@@ -0,0 +1,40 @@
...
+
+	# override original yaml file if differ
+	dif=$(diff -q "${yaml}" "${yaml}".tmp)
```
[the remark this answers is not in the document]
synarete: Indeed, I did not want to change timestamps if nothing changed. Re-thinking about it, I agree with you -- it confuses the naive reader. Will fix.

hack/yq-fixup-yamls.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
+# use yq to reformat yaml files to fit yamllint's syntax validity checks
+#
+# usage: 
+#   yq-fixup-yamls.sh file.yaml 
+#   yq-fixup-yamls.sh dirpath
+#
+YQ=${YQ:-$(command -v yq)}
+${YQ} --version > /dev/null
```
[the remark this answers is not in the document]
synarete: Sanity check for the YQ variable. Apparently, it is confusing. I will remove it.

hack/yq-fixup-yamls.sh
synarete: Do we have such cases? IMHO, it would be better to rename FOO.YAML --> foo.yaml.

Makefile
synarete: [yamllint](https://github.com/adrienverge/yamllint) is a python utility. YAMLLINT_CMD follows the same rules as KUBECTL_CMD and BUILDAH_CMD: expects the user to install them in his PATH

hack/yq-fixup-yamls.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
synarete: ACK

hack/yq-fixup-yamls.sh
phlogistonjohn: The issue is that it won't get caught. I'm not against preferring lowercase, I personally prefer it. So, while I won't insist on it I figured I would bring it up.

Makefile
phlogistonjohn: OK, thanks. Not for this PR, but we should probably think a little bit over how we handle our project dependencies/tools. I started an effort a couple of days ago but got bogged down "in the weeds". But I will probably revisit it soon.

hack/yq-fixup-yamls.sh
```diff
@@ -0,0 +1,29 @@
...
+#   yq-fixup-yamls.sh dirpath
+#
+set -e
+YQ=${YQ:-$(command -v yq)}
+
+_yaml_fixup_file() {
+	yaml="$1"
+
+	${YQ} eval "${yaml}" > "${yaml}".tmp
+	mv "${yaml}".tmp "${yaml}"
```
raghavendra-talur, line 15: There is an inplace option in yq if you prefer that approach.

synarete: Missed this option... using '--inplace' feels better (avoid leaving '*.tmp' files behind). Will change.

hack/yq-fixup-yamls.sh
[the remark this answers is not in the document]
phlogistonjohn: The use of `set -e` earlier in the file should take care of this now.

phlogistonjohn (dismissed):

phlogistonjohn (approved):

PR #134, opened 2022-01-20
Allow building image for any ARCH

synarete:
Do not assume specific hard-coded ARCH when building Docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@
 FROM docker.io/golang:1.17 as builder
 ARG GIT_VERSION="(unset)"
 ARG COMMIT_ID="(unset)"
+ARG ARCH=""
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
@@ -18,7 +19,7 @@ COPY controllers/ controllers/
 COPY internal/ internal/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on \
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=${ARCH} GO111MODULE=on \
     go build -a \
     -ldflags "-X main.Version=${GIT_VERSION} -X main.CommitID=${COMMIT_ID}" \
     -o manager main.go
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -14,21 +14,21 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 COMMIT_ID = $(shell git describe --abbrev=40 --always --dirty=+ 2>/dev/null)
 GIT_VERSION = $(shell git describe --match='v[0-9]*.[0-9].[0-9]' 2>/dev/null || echo "(unset)")
 
-CONFIG_KUST_DIR:=config/default
-CRD_KUST_DIR:=config/crd
-MGR_KUST_DIR:=config/manager
-KUSTOMIZE_DEFAULT_BASE:=../default
+CONFIG_KUST_DIR := config/default
```
phlogistonjohn, line 17: I approve of consistency, but I'd prefer it to be consistently without spaces. Would you object to that?

Dockerfile
```diff
@@ -2,6 +2,7 @@
 FROM docker.io/golang:1.17 as builder
 ARG GIT_VERSION="(unset)"
 ARG COMMIT_ID="(unset)"
+ARG ARCH="(unset)"
```
phlogistonjohn, line 5: The string "(unset)" has no special meaning and I used it to pass strings to the go binary so we'd know if the build never specified the GIT version and id. However I'm not sure it makes sense to set ARCH to "(unset)"  would it make more sense to default to empty string "" or "amd64" ?  Just so it defaults to something valid?

synarete: I was wondering about it as well. I would vote for empty string. Will fix.

Makefile
```diff
@@ -1,3 +1,6 @@
+# Alllow developer to override some defaults
+-include devel.mk
```
phlogistonjohn, line 2: Great! I learned something today. (Dash prefix to ignore errors when the files doesn't exist)

phlogistonjohn (approved):

PR #139, opened 2022-01-25
Install build tools

synarete:
Use helper script to install auxiliary build tools, using local GOBIN.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,10 @@ endif
 # Get current GOARCH
 GOARCH?=$(shell $(GO_CMD) env GOARCH)
 
+# Local (alternative) GOBIN for auxiliary build tools
+GOBIN_ALT:=$(CURDIR)/.bin/
+
+
 CONTAINER_BUILD_OPTS?=
 CONTAINER_CMD?=
 ifeq ($(CONTAINER_CMD),)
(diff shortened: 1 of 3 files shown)


hack/install-tools.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/sh
+# helper script to install build auxiliary tools in local directory
+#
+# usage:
+#   GOBIN=<dir> [TMPDIR=<tmpdir>] install-tools.sh --<tool-name>
+#
+set -e
+GO_CMD=${GO_CMD:-$(command -v go)}
+GOBIN=${GOBIN:-$(pwd)/bin}
+TMPDIR=${TMPDIR:-$(pwd)/tmp}
```
phlogistonjohn, line 10: Let's use `mktmp -d` in the functions that need a temporary dir.

Makefile
```diff
@@ -52,6 +52,10 @@ endif
 # Get current GOARCH
 GOARCH?=$(shell $(GO_CMD) env GOARCH)
 
+# Local (alternative) GOBIN for auxiliary build tools
+GOBIN_ALT:=$(CURDIR)/build/tools/bin/
```
phlogistonjohn, line 56: How about just `$(CURDIR)/.bin`  ?

Makefile
```diff
@@ -163,24 +167,25 @@ container-push:
 # download controller-gen if necessary
 controller-gen:
 ifeq (, $(shell command -v controller-gen ;))
-	@echo "controller-gen not found in PATH, checking in GOBIN ($(GOBIN))..."
-ifeq (, $(shell command -v $(GOBIN)/controller-gen ;))
-	$(GO_CMD) install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.6.2
-	@echo "controller-gen installed in GOBIN"
+	@echo "controller-gen not found in PATH, checking $(GOBIN_ALT)"
+ifeq (, $(shell command -v $(GOBIN_ALT)/controller-gen ;))
+	@GOBIN=$(GOBIN_ALT) GO_CMD=$(GO_CMD) hack/install-tools.sh --controller-gen
```
phlogistonjohn: Just to be on the safe side: `$(CURDIR)/hack/install-tools.sh` ?

phlogistonjohn (changes requested):
Really great start thanks! It's a lot simpler than than what I tried a few weeks ago and that makes it a lot more maintainable.

hack/install-tools.sh
synarete: ACK

Makefile
synarete: ACK

Makefile
synarete: ACK

synarete: (and we can even do it better ;)

phlogistonjohn (approved):
Very nice, thanks!

phlogistonjohn:
I'm leaving it open for a couple of days to give @raghavendra-talur a chance to look as we've discussed this topic regarding other projects in the past.

raghavendra-talur (approved):

PR #141, opened 2022-01-26
clean up how we run unit tests

phlogistonjohn:
Clean up the process of how we run unit tests, dropping the previous, not-really-used wrapper script.
Add some starter tests to internal/resources.
Make it simple to run unit tests and generate a coverage report in html.


internal/resources/configmap_test.go
```diff
@@ -0,0 +1,82 @@
+/*
```
synarete, line 1: Maybe it would be better to adopt the more common convention of SPDX-License-Identifier (a single comment at the top of each source file):

```// SPDX-License-Identifier: Apache-2.0```

spuiuk (approved):
ACK

internal/resources/configmap_test.go
phlogistonjohn: Sounds reasonable, but let's dedicate a separate PR (or more) to that effort.

raghavendra-talur (approved):

Issue #136, opened 2022-01-21
kustomization.yaml in developer guide is incomplete


spuiuk:
The contents for config/developer/kustomization.yaml should also include

resources:
- ../crd
- ../rbac
- ../manager-full

for it to work properly.

Without this, the command
make DEVEOPER=1 deploy will fail with the output

/home/sprabhu/go/bin/controller-gen "crd:trivialVersions=true,crdVersions=v1" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
cd config/developer && /home/sprabhu/go/bin/kustomize edit set image controller=quay.io/spuiuk/smbshare_devel:test
/home/sprabhu/go/bin/kustomize build config/developer | kubectl apply -f -
Error: merging from generator &{0xc0004b6360 {  map[] map[]} {{system controller-cfg merge {[SAMBA_OP_SAMBA_DEBUG_LEVEL=10 SAMBA_OP_CLUSTER_SUPPORT=ctdb-is-experimental] [] [] } <nil>}}}: id resid.ResId{Gvk:resid.Gvk{Group:"", Version:"v1", Kind:"ConfigMap", isClusterScoped:false}, Name:"controller-cfg", Namespace:"system"} does not exist; cannot merge or replace

spuiuk:
You also need to add 
namespace: samba-operator-system
namePrefix: samba-operator-

to the kustomization.yaml file for it to work properly. Else it installs the operator in the namespace system instead.

PR #142, opened 2022-01-26
Improve developer workflow in makefile and docs

phlogistonjohn:
Previously, the docs didn't explain that the example YAML snippets were to be added to a stock kustomization.yaml, nor did it explain that the makefile could help create this kustomization.yaml.
Now we  can easily initialize the developer dir with `make developer-dir DEVELOPER=1` and we improve the docs to match.

Fixes: #136

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,10 +122,26 @@ deploy: manifests kustomize set-image
 delete-deploy: manifests kustomize
 	$(KUSTOMIZE) build $(CONFIG_KUST_DIR) | $(KUBECTL_CMD) delete -f -
 
-%/kustomization.yaml: $(KUSTOMIZE)
+# the bar symbol below is an order only prerequisite
+#  https://www.gnu.org/software/make/manual/make.html#index-order_002donly-prerequisites
+# this is needed because kustomize is phony but we really do not want to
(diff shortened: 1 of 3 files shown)


synarete:
LGTM. 
While reviewing, I noticed two minor issues, which worth squash into this PR: 
https://github.com/synarete/samba-operator/commit/2fdedbcda3ce2d788333e34d417debb0db622325
https://github.com/synarete/samba-operator/commit/846a99f615f003cec88a0610630c124d907d9937

spuiuk (dismissed):
Looks Good.

phlogistonjohn:
PR updated, PTAL.  @spuiuk @synarete

spuiuk (dismissed):
ACK

phlogistonjohn:
I think the CI is actually failing due to this PR's changes. It is certainly going to be part of the makefile changes, not the doc text but I need to hunt down what it is because the changes should have been minor.

phlogistonjohn:
PR updated. I finally figured out the issue wrt why the makefile changes were breaking the CI. Details are in a comment in the change. When you have a free moment, PTAL @spuiuk @synarete

spuiuk (approved):
ACK

PR #143, opened 2022-01-27
makefile: reformat yaml upon 'make manifests'

### LINKED ISSUE #140: YAML files "churn" when executing make commands
I think this is fallout from #131 

When I run commands such as `make manifests` many of the YAML files show as changed to git. However, I've not touched the sources. I think we should do one of the following:

* Configure our yaml generating tools to conform to the new styling
* Configure the make file to automatically format the output of `make manifests` and the like
* Change our formatting/linting rules to conform to the default output of the tools used for `make manifests` and the like.

synarete:
Tools like 'controller-gen' generates output YAML files which does not
adhere the strict format of 'yamllint'. Re-run yamls re-formatter script
and the end of 'make manifests' rule.

Fixes #140


phlogistonjohn (approved):
Looks good, thanks.

raghavendra-talur (approved):

PR #144, opened 2022-01-31
OpenShift deploy

synarete:
Enable deployment on OpenShift cluster, where smbd container can not run with UID=0, and can not bind to privileged ports. Pass explicit service and ambd ports via operator-config.

Tested over OpenShift 4.9

### DIFF
--- a/internal/resources/pods.go
+++ b/internal/resources/pods.go
@@ -436,27 +436,28 @@ func buildSmbdCtr(
 	env []corev1.EnvVar,
 	vols []volMount) corev1.Container {
 	// ---
+	portnum := planner.GlobalConfig.SmbdPort
 	return corev1.Container{
 		Image: planner.GlobalConfig.SmbdContainerImage,
 		Name:  planner.GlobalConfig.SmbdContainerName,
 		Args:  planner.runDaemonArgs("smbd"),
 		Env:   env,
 		Ports: []corev1.ContainerPort{{
-			ContainerPort: 445,
+			ContainerPort: int32(portnum),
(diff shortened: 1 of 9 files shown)


internal/resources/pods.go
```diff
@@ -668,6 +658,13 @@ func defaultPodEnv(planner *sharePlanner) []corev1.EnvVar {
 	return env
 }
 
+func defaultPodSpec(planner *sharePlanner, sharens *bool) *corev1.PodSpec {
```
phlogistonjohn, line 661: Why is it returning a pointer here if all the other uses have to first dereference it? Wouldn't it be simpler to just return `corev1.PodSpec` here?

phlogistonjohn (changes requested):
Looks like a great start. I have a few minor things that need attention inline.

I'm not 100% sure if you consider the PR fully done, because I don't see the new kustomization dir being used. I'd like to see at least some new developer doccs that show how to consume the additional YAMLs and set the service account int the operator config. But if you'd prefer to do that in a follow up PR just say so, and I'll treat this one as complete/reviewable.

internal/resources/pods.go
synarete: You are right, will fix.

As a rule of thumb, I prefer to avoid aggregate return in C, and thus also in Golang (avoid redundant copies), although I am not sure if Go compiler will simply optimize the case above. In all odds, in the context of this patch it is just confusing. Will fix.

phlogistonjohn (approved):
LGTM, thanks!

PR #148, opened 2022-02-10
OpenShift deploy [2]

synarete:
When deploying over OpenShift, need to setup SCC 'anyuid' to samba-server pod. 
Configure the appropriate ServiceAccount, Role and RoleBinding from within samba-operator.

### DIFF
--- a/config/manager-full/auth_proxy_patch.yaml
+++ b/config/manager-full/auth_proxy_patch.yaml
@@ -29,3 +29,16 @@ spec:
           args:
             - "--metrics-addr=127.0.0.1:8080"
             - "--enable-leader-election"
+          env:
+            - name: SAMBA_OP_POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: SAMBA_OP_POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: SAMBA_OP_POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -65,6 +65,19 @@ rules:
     verbs:
       - get
       - list
+      - watch
+  - apiGroups:
+      - ""
+    resources:
+      - serviceaccounts
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - patch
+      - update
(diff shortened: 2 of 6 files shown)


phlogistonjohn (commented):
Sorry for taking a long time to respond to this. A couple of high-level questions:
* Because the reconciliation happens in the SmbCommonConfig, what should happen for SmbShares that don't explicitly have a common config. Today, a common config is not required for an SmbShare and it "just inherits" some baked in defaults
* Do we need a way to override the detected state manually, say thru a specifc config var, mainly for testing and as an "escape hatch" in case autodetection is doing something wrong?
* Perhaps I'm missing something but how does the new scc support apply to the smb server pods the operator creates? Is that meant to come in a later patch series?

synarete:
1) SCC should be configured once (and only once) for each samba-operator-system, hence it is triggered by SmbCommonConfig (same as network-specific settings, or, in the future, namespace-labeling for Prometheus metrics scraping). An attempt to create SmbShares without SmbCommonConfig over OpenShift will yield pod failure due to missing permissions. IMHO it is a valid requirement from OpenShift users to create SmbCommonConfig as part of their smb setup.

synarete:
2) The override option is an interesting question. It is a fine balance between developer's flexibility and code maintenance/complexity.  In need to think about it a bit more, and we should probably also discuss it f2f.

synarete:
3) SCC (and its related Role and RoleBinding) is associated to smb-server pods via the ServiceAccount. Indeed, it is somewhat hidden and hard-to-spot binding: 

https://github.com/synarete/samba-operator/blob/ss-openshift-deploy2/internal/resources/pods.go#L670

https://github.com/synarete/samba-operator/blob/ss-openshift-deploy2/internal/resources/scc.go#L228

synarete:
> Any news? It would be great to not have to configure the rolebindings manually for every namespace a share lives in.

@ibotty OpenShift patches are sill under review (see #216). Unfortunately, it did not make make it to `v0.2` but most likely will be ready for next release. I build and test a custom samba-operator over openshift (4.8) on a daily basis. Feel free to use it and let me know if you have any problems: `quay.io/ssharon/sink:latest`

phlogistonjohn:
I'm making an "executive decision" to close this PR. AFAIK, the feature set here was superseded by other PRs that are now merged and there's no expectation that this PR will ever be merged. If I'm wrong, feel free to reopen it.

PR #149, opened 2022-02-21
hack: bump kustomize version to v4.5.2

synarete:
Latest stable version of kustomize fixes previous issues which prevented
from installing of 'kustomize' helper build-tool via 'go install'. Using
newer version we no longer need to use 'go get' in temporary directory,
and for users of go1.17 there isn't any deprecation notice.


phlogistonjohn (approved):
Looks good, glad to see they cleaned some things up.

PR #152, opened 2022-03-01
resources: always share process namespace within pod

synarete:
Always set ShareProcessNamespace to 'true' in pod's spec, so that
different containers within same pod will see each others process pids.
This functionality is required for proper operation of smb tools (e.g.,
smbstatus) which expects the pid of 'smbd' to be a valid value within
their process namespace.


phlogistonjohn (approved):
LGTM, thanks!

PR #155, opened 2022-03-04
Few cleanup fixes

anoopcs9:
* Removes an extra forward slash **_/_**
* Replace **_bases_** with **_resources_** for _kustomize_. **_bases_** field was [deprecated](https://kubectl.docs.kubernetes.io/blog/2019/06/18/v2.1.0/#resources-expanded-bases-deprecated) long time back in favour of **_resources_**.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ endif
 GOARCH?=$(shell $(GO_CMD) env GOARCH)
 
 # Local (alternative) GOBIN for auxiliary build tools
-GOBIN_ALT:=$(CURDIR)/.bin/
+GOBIN_ALT:=$(CURDIR)/.bin
 
 
 CONTAINER_BUILD_OPTS?=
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -53,7 +53,7 @@ endif
 GOARCH?=$(shell $(GO_CMD) env GOARCH)
 
 # Local (alternative) GOBIN for auxiliary build tools
-GOBIN_ALT:=$(CURDIR)/.bin/
+GOBIN_ALT:=$(CURDIR)/.bin
```
phlogistonjohn, line 56: This change seems unrelated to the kustomize/kubectl deprecation of "bases". If this is a useful change I'd like to see it split out and described, thanks.

phlogistonjohn (changes requested):

Makefile
anoopcs9: Done.

phlogistonjohn (approved):
LGTM, thanks!

PR #158, opened 2022-03-06
centosci: Enable clustered test cases

anoopcs9:


tests/integration/smb_share_test.go
```diff
@@ -203,7 +203,7 @@ func (s *SmbShareWithDNSSuite) TestShareAccessByDomainName() {
 	// test that the IP in ad dns matches the service
 	ctx, cancel := context.WithDeadline(
 		context.TODO(),
-		time.Now().Add(30*time.Second))
+		time.Now().Add(60*time.Second))
```
spuiuk, line 206: All the changes look good and I have only one small request. 
Instead of using a number here directly, can you use a variable defined as a global in tests/integration/util_test.go. This is far more easily maintainable.

spuiuk (changes requested):
I like these changes but have one minor request.

tests/integration/smb_share_test.go
anoopcs9: These changed are now included with PR #162, thanks.

tests/integration/util_test.go
```diff
@@ -10,6 +10,7 @@ import (
 )
 
 var (
+	waitForIpTime    = 120 * time.Second
```
phlogistonjohn, line 13: Your commit says "1m to be on safe side" but this is two minutes, unless I'm really tired. :-)
I don't have a strong preference but I'd like to see the commit and code agree.

phlogistonjohn (commented):
I have a small nitpick with the commit message/code agreeing. Beyond that, what in your opinion, @anoopcs9  does this PR need to be reviewable?

tests/integration/util_test.go
anoopcs9: Fixed and updated in #162

spuiuk (dismissed):

spuiuk (approved):
Acked-by: Sachin Prabhu <sprabhu@redhat.com>

PR #161, opened 2022-03-11
Add support for Spec.Storage.Pvc.Path component

spuiuk:
This allows users to specify the path within a pvc which should be exported by samba. Includes an integration test which tests this new feature.

### DIFF
--- a/api/v1alpha1/smbshare_types.go
+++ b/api/v1alpha1/smbshare_types.go
@@ -86,6 +86,10 @@ type SmbSharePvcSpec struct {
 	// Behaves similar to the embedded PVC spec for pods.
 	// +optional
 	Spec *corev1.PersistentVolumeClaimSpec `json:"spec,omitempty"`
+
+	// Path within the PVC which should be exported.
+	// +optional
+	Path string `json:"path,omitempty"`
 }
 
 // SmbShareScalingSpec defines scaling parameters for a share.
--- a/api/v1alpha1/smbshare_types.go
+++ b/api/v1alpha1/smbshare_types.go
@@ -88,6 +88,7 @@ type SmbSharePvcSpec struct {
 	Spec *corev1.PersistentVolumeClaimSpec `json:"spec,omitempty"`
 
 	// Path within the PVC which should be exported.
+	// +kubebuilder:validation:Pattern=`^[^\/]+$`
 	// +optional
 	Path string `json:"path,omitempty"`
 }
--- a/tests/integration/mount_path_test.go
+++ b/tests/integration/mount_path_test.go
@@ -1,0 +1,139 @@
+// +build integration
+
+package integration
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/stretchr/testify/suite"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/samba-in-kubernetes/samba-operator/tests/utils/kube"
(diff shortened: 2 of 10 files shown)


spuiuk:
Setting to draft for now since the path update test fails.

internal/resources/planner.go
```diff
@@ -310,7 +310,7 @@ func (sp *sharePlanner) update() (changed bool, err error) {
 			changed = true
 		}
 	}
-	return
+	return changed, nil
```
phlogistonjohn, line 313: Is this needed? The function uses named return values [1]. Did you see a bug with this line?

1 -  https://go.dev/tour/basics/7

tests/files/client-test-pod-userpvc.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: smbclient-withpvc
```
phlogistonjohn, line 4: Is this used for anything now?

tests/integration/mount_path_test.go
```diff
@@ -0,0 +1,139 @@
...
+	smbshareSetupSources    []kube.FileSource
+	smbshareSources         []kube.FileSource
+	smbShareSetupResource   types.NamespacedName
+	setupServerLabelPattern string
+	smbShareResource        types.NamespacedName
+	serverLabelPattern      string
+	tc                      *kube.TestClient
+}
+
+func (s *MountPathSuite) getPodIP(labelPattern string) (string, error) {
```
phlogistonjohn, line 31: In the future we probably want to decrease the number of tests the rely primarily on reaching the smb share via the pods IP address. It's not an issue for now, but something to keep in mind for later.

phlogistonjohn (changes requested):

api/v1alpha1/smbshare_types.go
```diff
@@ -88,6 +88,7 @@ type SmbSharePvcSpec struct {
 	Spec *corev1.PersistentVolumeClaimSpec `json:"spec,omitempty"`
 
 	// Path within the PVC which should be exported
+	// +kubebuilder:validation:Pattern=`^\/[^\/]+$`
```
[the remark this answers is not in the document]
spuiuk: An old habit. I've changed it to your suggestion in the latest version.

internal/resources/planner.go
spuiuk: I wasn't aware of named return values and to a person used to looking at C code, seemed wrong. I was just investigating the code around when I was confused about this. I've deleted the change but we should consider avoiding this particular feature to avoid such confusion. :)

tests/files/client-test-pod-userpvc.yaml
spuiuk: You are right. This is no longer used and just something I missed when updating the code.

tests/integration/mount_path_test.go
spuiuk: I simply used the ip address as that showed up on the first example I looked at. I have added a new patch which removes the use of the ip address. I hope you don't mind me keeping this as a separate patch as I can use these as a reference for when I add more tests.

phlogistonjohn: I think that'll be fine.

phlogistonjohn:
As discussed over voice the planner refactor PR has been merged. You'll need to manually rebase and resolve the conflicts that arise from the changed planner code.

phlogistonjohn (approved):
LGTM, thanks!

PR #163, opened 2022-03-14
Improvements to CentOS CI test scripts

anoopcs9:
* Use alternate approach for loading container images onto minikube cluster
* Increase deployment wait timeout for rook.


phlogistonjohn (approved):
thanks!

PR #164, opened 2022-03-15
Metrics exporter container within smbd pod

synarete:
Provide the mechanism to execute samba metrics exporter as container along side smbd, within same pod. 

The code of smbmetrics is now developed in a separate repository:
https://github.com/samba-in-kubernetes/smbmetrics

### DIFF
--- a/internal/conf/config.go
+++ b/internal/conf/config.go
@@ -10,16 +10,18 @@ import (
 
 // DefaultOperatorConfig holds the default values of OperatorConfig.
 var DefaultOperatorConfig = OperatorConfig{
-	SmbdContainerImage:     "quay.io/samba.org/samba-server:latest",
-	SvcWatchContainerImage: "quay.io/samba.org/svcwatch:latest",
-	SmbdContainerName:      "samba",
-	WinbindContainerName:   "wb",
-	WorkingNamespace:       "",
-	SambaDebugLevel:        "",
-	StatePVCSize:           "1Gi",
-	ClusterSupport:         "",
-	SmbServicePort:         445,
-	SmbdPort:               445,
+	SmbdContainerImage:        "quay.io/samba.org/samba-server:latest",
+	SmbdMetricsContainerImage: "quay.io/samba.org/samba-metrics:latest",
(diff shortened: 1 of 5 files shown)


internal/conf/config.go
```diff
@@ -10,16 +10,18 @@ import (
...
+	SvcWatchContainerImage:    "quay.io/samba.org/svcwatch:latest",
+	SmbdContainerName:         "samba",
+	WinbindContainerName:      "wb",
+	WorkingNamespace:          "",
+	SambaDebugLevel:           "",
+	StatePVCSize:              "1Gi",
+	ClusterSupport:            "",
+	SmbServicePort:            445,
+	SmbdPort:                  445,
+	WithMetricsExporter:       false,
```
phlogistonjohn, line 24: I have a general issue with booleans when used in in certain contexts like configuration or other kinds of persistent data. I'd like to suggest a tweak to this to take a string and tweak the name to something like "MetricsExporterMode" accept values such as empty string (""), "enabled", "disabled". Then in the future, if we need different approaches for the exporter, say "debug" hypothetically we don't get an explosion of fields like you would if you needed to keep adding various on/off booleans for features.  Basically, I want an "enum" but since we can't directly use an enum - we have an enumerated set of string values. If this isn't clear, feel free to ping me and I'll try to elaborate more.

It is acceptable to fail the operator if they feed in a unknown value.

phlogistonjohn (changes requested):
The core logic looks pretty good, so this is a pretty nice start.

I have a suggestion for how we approach config items.

internal/conf/config.go
synarete: ACK. Fixed (+rebase over latest master)

phlogistonjohn (approved):
LGTM. Thanks a bunch for working on this!

I highly encourage you to make use of github's feature for re-requesting reviews ( there's a little "recycle" icon next to my name in the reviewrs box).
I, as well as others, tend to be driven by new mail but github doesn't apparently send email notifications on force pushes. I'm pretty sure it does if you re-request a reivew. Also feel free to @-ping me in a comment if the pr hasn't been looked at in a couple of days.

PR #172, opened 2022-03-24
Ensure the smbclient pod is up and running before tests

spuiuk:
There are many tests which rely on the smbclient pod already up and running.
This is a problem if individual tests are run.

We add a new function createSMBClientIfMissing() to tests/integration/util_test.go to run the smbclient if it isn't already running.

At the moment, we do not ensure that the running smbclient is delete once the tests have completed running. This set of patches do not address that at the moment. I will post an additional set of patches to fix this condition later.


phlogistonjohn:
> At the moment, we do not ensure that the running smbclient is delete once the tests have completed running. This set of patches do not address that at the moment. I will post an additional set of patches to fix this condition later.

I do not want the smbclient pod to be deleted at the end of the test run. When debugging a test failure its highly inconvenient to have to redeploy all resources needed (the operator, the smbclient pod, etc) esp. when running the tests locally. Cleaning up the resoouces after the tests complete is often as simple as deleting the namespace if you really want to restart clean. If you still really want to do this I ask that this cleanup behavior be fully optional and easy to deactivate.

phlogistonjohn (approved):
Changes LGTM.

anoopcs9 (approved):

PR #179, opened 2022-04-06
resources: add a volKeeper type for wrangling volumes

phlogistonjohn:
The current implementation of pod construction relies on managing many
slices of volMount objects. Not only is this a bit verbose but it also
relies on a trick that the static checkers don't like. In order to
make things a bit less "loose" and more regular the very simple
volKeeper type is added as a wrapper around a slice of volMounts.

The major pod assembly functions are then converted to use volKeepers.

This work is a precursor to additional patches that will add new containers to the pods
that will watch for configuration changes.

### DIFF
--- a/internal/resources/pods.go
+++ b/internal/resources/pods.go
@@ -52,33 +52,30 @@ func buildADPodSpec(
 	_ *conf.OperatorConfig,
 	pvcName string) corev1.PodSpec {
 	// ---
-	volumes := []volMount{}
-	smbAllVols := []volMount{}
+	volumes := newVolKeeper()
+	smbAllVols := newVolKeeper()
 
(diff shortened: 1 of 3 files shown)


synarete (approved):
Big improvement.
I also tested this patch-set over OpenShift with smbmetrics enabled (rebased over master). All good.
LGTM.

phlogistonjohn:
Some suggestions for the future: 
* If you're reviewing the PR and it's taking a long time, put a comment on the PR and other folks will know
* If you're reviewing and you strongly want the PR not to be merged until you're done, put the `do-not-merge` label on the PR (with an accompanying comment :-) )
* In this case where the PR has been merged, keep reviewing and if you do find something you think needs to be changed file new issues.

Thanks for looking!

PR #180, opened 2022-04-11
tests: append local-tools bin dir to PATH

synarete:
Some developers uses 'kustomize' from local '.bin' directory, and
therefore it should be appended to search PATH variable as part of
tests-execution wrapper script.


phlogistonjohn (approved):
Seems OK. Maybe someday we'll find our way out of kustomize konfusion. ;-)

PR #185, opened 2022-04-15
tests: a script to dump various logs/state info 

phlogistonjohn:
When a test fails, especially if that test is a CI job running on remote systems, it is difficult to debug/diagnose issues. These changes add a script that can be run after the tests fail in order to grab some information about the state of the cluster, including operator logs.
It may not be sufficient to debug all issues/failures but it could be a helpful first step.

### DIFF
--- a/tests/centosci/sink-clustered-deployment.sh
+++ b/tests/centosci/sink-clustered-deployment.sh
@@ -34,7 +34,7 @@ IMG="${CI_IMG_OP}" make test
 
 # Run integration tests
 SMBOP_TEST_CLUSTERED=1 SMBOP_TEST_EXPECT_MANAGER_IMG="${CI_IMG_OP}" \
-	./tests/test.sh
+	./tests/test.sh || (./tests/post-test-info.sh; exit 1)
 
 teardown_op
 
--- a/tests/centosci/sink-clustered-deployment.sh
+++ b/tests/centosci/sink-clustered-deployment.sh
@@ -13,7 +13,7 @@ setup_minikube
 
 deploy_rook
 
-image_pull ${CI_IMG_REGISTRY} "docker.io" "golang:1.17"
+image_pull "${CI_IMG_REGISTRY}" "docker.io" "golang:1.17"
 
 # Build and push operator image to local CI registry
 IMG="${CI_IMG_OP}" make image-build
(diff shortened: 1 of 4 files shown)


tests/centosci/sink-clustered-deployment.sh
```diff
@@ -34,7 +34,7 @@ IMG="${CI_IMG_OP}" make test
 
 # Run integration tests
 SMBOP_TEST_CLUSTERED=1 SMBOP_TEST_EXPECT_MANAGER_IMG="${CI_IMG_OP}" \
-	./tests/test.sh
+	./tests/test.sh || ./tests/post-test-info.sh
```
anoopcs9, line 37: But this would make overall test to be marked as successful. Don't we need to report CI run as failed?

anoopcs9, line 37: But this would make overall test to be marked as successful. Don't we need to report CI run as failed?

phlogistonjohn: Good catch, will fix.

phlogistonjohn: Updated

tests/centosci/sink-clustered-deployment.sh
```diff
@@ -34,7 +34,7 @@ IMG="${CI_IMG_OP}" make test
 
 # Run integration tests
 SMBOP_TEST_CLUSTERED=1 SMBOP_TEST_EXPECT_MANAGER_IMG="${CI_IMG_OP}" \
-	./tests/test.sh
+	./tests/test.sh || (res=$? ; ./tests/post-test-info.sh; exit "$res")
```
anoopcs9, line 37: How about the following change?

```suggestion
	./tests/test.sh || (./tests/post-test-info.sh && exit 1)
```

anoopcs9 (commented):
Looks OK. A small suggestion though

tests/centosci/sink-clustered-deployment.sh
phlogistonjohn: It's fine, but I figured I'd go the extra mile and preserve the original error code. Please confirm if you'd like the "exit 1" version and I'll make the change.

phlogistonjohn: done

anoopcs9 (approved):

PR #188, opened 2022-04-25
tests: patch dns-operator when running over OpenShift

synarete:
When running over OpenShift need to patch the CRD of dns-operator in
order order to configure extra DNS entry for testing. Using 'jq' utility
to simplify JSON parsing.

See "DNS Operator in OpenShift Container Platform" in OpenShift's
documentations for examples.

### DIFF
--- a/tests/test-deploy-ad-server.sh
+++ b/tests/test-deploy-ad-server.sh
@@ -6,12 +6,21 @@ BASE_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
 DEPLOYMENT_YAML="${BASE_DIR}/tests/files/samba-ad-server-deployment.yml"
 DEPLOYMENT_NAME="samba-ad-server"
 COREDNS_SNIPPET="${BASE_DIR}/tests/files/coredns-snippet.template"
+OPENSHIFT_DNS_SNIPPET="${BASE_DIR}/tests/files/openshift-dns-snippet.template"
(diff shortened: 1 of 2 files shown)


tests/test-deploy-ad-server.sh
```diff
@@ -65,12 +74,31 @@ echo
...
+	PATCHFILE=$(mktemp).json
+	sed -e "s/AD_SERVER_IP/${AD_POD_IP}/g" \
+		${OPENSHIFT_DNS_SNIPPET} > "${PATCHFILE}"
+
+	${KUBECTL_CMD} patch dns.operator default \
+		--type=merge --patch-file "${PATCHFILE}"
+
+	[ $? -eq 0 ] || _error "Failed patching dns-operator with $PATCHFILE"
+	unlink "${PATCHFILE}"
+	exit 0
```
phlogistonjohn, line 99: Nitpick: I missed this the first readthrough and for a moment thought the code was going to run this if-block AND the regular non-openshift approach to patching the coredns config.  I won't ask for it in this PR, but next time we need to do non-trivial work to this script it might be time to start using (more) functions,

phlogistonjohn (approved):
Based on what little I know about openshift's specific approach to coredns configuration this  seems good to me.

phlogistonjohn:
I'll give the centos ci one more chance. :-)
If it doesn't work again I can just merge this since the CI is not really testing this part of the script.

phlogistonjohn (approved):

PR #194, opened 2022-05-11
tests: ensure samba-metrics sidecar is responsive

synarete:
Issue a curl from samba container to samba-metrics sidecar container in
order to verify samba-metrics is responsive with valid output. Execute
the curl command within the pod itself.

Enabled only if 'samba-metrics' container exists in pod.

### DIFF
--- a/tests/integration/smb_share_test.go
+++ b/tests/integration/smb_share_test.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/stretchr/testify/suite"
@@ -76,6 +77,14 @@ func (s *SmbShareSuite) getPodFetchOptions() kube.PodFetchOptions {
 }
 
 func (s *SmbShareSuite) getPodIP() (string, error) {
+	pod, err := s.getReadyPod()
+	if err != nil {
+		return "", err
+	}
+	return pod.Status.PodIP, nil
+}
+
+func (s *SmbShareSuite) getReadyPod() (*corev1.Pod, error) {
 	l := fmt.Sprintf(
 		"samba-operator.samba.org/service=%s", s.smbShareResource.Name)
 	pods, err := s.tc.FetchPods(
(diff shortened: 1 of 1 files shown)


tests/integration/smb_share_test.go
```diff
@@ -273,6 +282,57 @@ func (s *SmbShareWithExternalNetSuite) TestServiceIsLoadBalancer() {
 	)
 }
 
+func (s *SmbShareSuite) TestMetricsOnPod() {
+	pod, cont, err := s.getMetricsContainer()
+	s.Require().NoError(err)
+	if cont == nil {
+		return // OK, running without metrics; ignored
```
phlogistonjohn, line 289: Please change this to a test skip. Example: `s.T().Skipf("no metrics container present")`

phlogistonjohn (changes requested):
Body of the test looks OK to me. I'd just like to use an explicit test skip to distinguish from the success case when the test actually runs.

tests/integration/smb_share_test.go
synarete: ACK

synarete:
@anoopcs9 @phlogistonjohn CI testing keep on failing on this PR as well as my other (trivial) PR#196. I  re-checked both on my setup (with minikube) and PASS with not errors. Any idea what could cause those failures?

anoopcs9:
> @anoopcs9 @phlogistonjohn CI testing keep on failing on this PR as well as my other (trivial) PR#196. I re-checked both on my setup (with minikube) and PASS with not errors. Any idea what could cause those failures?

I'm looking into CentOS CI test failures..I'm able to reproduce where I am seeing following errors from Ceph cluster which prevents it to come up HEALTHY:

`MON_CLOCK_SKEW`
`MON_DISK_LOW`

clcok skew is surprising but I think I figured out a way to get around `MON_DISK_LOW` error.

anoopcs9:
> > @anoopcs9 @phlogistonjohn CI testing keep on failing on this PR as well as my other (trivial) PR#196. I re-checked both on my setup (with minikube) and PASS with not errors. Any idea what could cause those failures?
> 
> I'm looking into CentOS CI test failures..I'm able to reproduce where I am seeing following errors from Ceph cluster which prevents it to come up HEALTHY:
> 
> `MON_CLOCK_SKEW` `MON_DISK_LOW`
> 
> clock skew is surprising but I think I figured out a way to get around `MON_DISK_LOW` error.

#197 should fix rook-ceph errors from CentOS CI test runs.

synarete:
> > > @anoopcs9 @phlogistonjohn CI testing keep on failing on this PR as well as my other (trivial) PR#196. I re-checked both on my setup (with minikube) and PASS with not errors. Any idea what could cause those failures?
> > 
> > 
> > I'm looking into CentOS CI test failures..I'm able to reproduce where I am seeing following errors from Ceph cluster which prevents it to come up HEALTHY:
> > `MON_CLOCK_SKEW` `MON_DISK_LOW`
> > clock skew is surprising but I think I figured out a way to get around `MON_DISK_LOW` error.
> 
> #197 should fix rook-ceph errors from CentOS CI test runs.

Cool! many thanks @anoopcs9

phlogistonjohn (approved):

PR #200, opened 2022-05-23
Set pvc permissions

spuiuk:
Patch to set the permissions on the mount point of the smbshare to 0700 by default using the initialize-share-perms permissiona handler. Accompanying this change is the integration test to check if the permissions handler was run against the mount path.


phlogistonjohn (changes requested):
Maybe this is kinda silly but I see that the first commit is labeled "resources:" but doesn't touch the resources package. Rather the change touches _both_ planner and smbcc. Ideally, each pkg would have its own commit. At the very least I'd like the commit topic changed to match the code changes. Thanks.

phlogistonjohn (approved):

spuiuk:
Error: 
         	Error:      	Received unexpected error:
        	            	too many pods match selector (opts: {Namespace:samba-operator-system LabelSelector:samba-operator.samba.org/service=tshare1 MaxFound:1 MinFound:0}; pods: [])

Looks like it is waiting on some cleanup. Unlikely to be caused by permission changes. The corresponding tests on the CentosCI have passed. 

Let me try kicking this off again.

spuiuk:
All clear now :)

phlogistonjohn:
OK, cool. I guess I just had bad luck.

PR #202, opened 2022-05-24
Go1.18 fixes

synarete:
Fixed various errors from 'make check' with go-1.18.2 (Linux, amd64, RHEL8).


controllers/smbshare_controller.go
```diff
@@ -72,9 +73,7 @@ func (r *SmbShareReconciler) Reconcile(
 }
 
 func (r *SmbShareReconciler) setRecorder(mgr ctrl.Manager) {
-	if r.recorder == nil {
-		r.recorder = mgr.GetEventRecorderFor("smbshare-controller")
-	}
+	r.recorder = mgr.GetEventRecorderFor("smbshare-controller")
```
phlogistonjohn, line 76: Just for my education: the linter thinks that recorder is always nil here?

tests/utils/kube/status.go
```diff
@@ -119,7 +119,7 @@ func WaitForAnyPodReady(
 		tc:        tc,
 		fetchOpts: fo,
 	}
-	pp.Cond = func() (bool, error) { return pp.checkReady(ctx) }
+	pp.Prober.Cond = func() (bool, error) { return pp.checkReady(ctx) }
```
phlogistonjohn, line 122: Has the type changed with the new go version or was this a bug all along?

controllers/smbshare_controller.go
```diff
@@ -62,7 +62,8 @@ func (r *SmbShareReconciler) Reconcile(
 	reqLogger.Info("Reconciling SmbShare")
 
 	smbShareManager := resources.NewSmbShareManager(
-		r, r.Scheme(), r.recorder, reqLogger)
+		r, r.Scheme(), r.recorder, reqLogger) // nolint:typecheck
```
phlogistonjohn, line 65: Just for my education: what kind of linter error was this generating without the suppresion comment?

phlogistonjohn (commented):
Typo in commit msg: "gi:build" should be "go:build" IIUC.

controllers/smbshare_controller.go
synarete: The actual output from the linter is: 
```
controllers/smbshare_controller.go:75:19: invalid operation: cannot compare r.recorder == nil (operator == not defined on untyped nil) (typecheck)
```

because  record.EventRecorder is interface

tests/utils/kube/status.go
synarete: It is not a bug, but rather a picky linter. I could not find out why in go1.18 they prefer this style.

controllers/smbshare_controller.go
synarete: ```r.Scheme undefined (type *SmbShareReconciler has no field or method Scheme)```

phlogistonjohn: That's weird. Probably a bug in the linter - or else this code shouldn't compile :-)

controllers/smbshare_controller.go
phlogistonjohn: Ah, OK.  At first I was a little concerned about this change because it is removing actual logic. But looking at the function being called `setRecorder` its pretty weird that it would have ignored the "set" if recorder was already assigned anyway. I'm now fine with the change.

controllers/smbshare_controller.go
synarete: Indeed -- probably linter bug

PR #207, opened 2022-06-06
resources: pass smbdVols to buildEnsureShareCtr

synarete:
Commit b140141 ("resources: create ensure-share-paths init container")
introduced a new init container, which changes the mode of share's
directory. Unfortunately, in buildADPodSpec we pass the wrong volumes
slice (a slice which does _not_ have the share volume). This bug is
evident only when running with smbmetrics enabled, on OpenShift.


synarete:
> Just to clarify, are you saying that the metrics container needs the share data volume? If so, it's something I'd like to avoid, but won't block - just track for future improvement. Otherwise change looks fine and I'm willing to approve once we've discussed this.

It is not for the metrics container. It is for the ensure-share-create (a.k.a chmod) container.

PR #210, opened 2022-06-06
config: default to latest samba operator image

phlogistonjohn:
Since using the Operator SDK to create the initial kustomization files
we've defaulted to 'controller:latest' in the base yamls. However, this
means one *must* use kustomize to set the image to something. Typically,
we've been doing this using the Makefile.
However, now that we've well established that the canonical image for
the operator is at 'quay.io/samba.org/samba-operator' there's really no
need to require the image to be overridden. Update the manager base
YAML to default to the canonical operator image repo - and update the
makefile to match.

Note: if you do have an existing   kustomization.yaml  referring to 'controller' you need to update that manually

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,7 +144,7 @@ endif
 .PHONY: developer-dir
 
 set-image: kustomize $(MGR_KUST_DIR)/kustomization.yaml
-	cd $(MGR_KUST_DIR) && $(KUSTOMIZE) edit set image controller=$(IMG)
+	cd $(MGR_KUST_DIR) && $(KUSTOMIZE) edit set image quay.io/samba.org/samba-operator=$(IMG)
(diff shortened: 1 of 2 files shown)


spuiuk (approved):
Acked-by: Sachin Prabhu <sprabhu@redhat.com>

spuiuk:
At the moment, the latest tag points to the latest devel build. Should this continue to point to the latest devel version or should it point to the latest released version?

phlogistonjohn:
> At the moment, the latest tag points to the latest devel build. Should this continue to point to the latest devel version or should it point to the latest released version?

I think its most common for a "young" project to have latest track development code. If the project matures we can reevaluate, but for now I think this is the right choice.

PR #213, opened 2022-06-08
Image pull policy

synarete:
Make default ImagePullPolicy=IfNotPresent in order to speed-up pod's start-up time which in turn reduces the chance of tests-failure due to timeouts.

### DIFF
--- a/internal/conf/config.go
+++ b/internal/conf/config.go
@@ -22,6 +22,7 @@ var DefaultOperatorConfig = OperatorConfig{
 	SmbServicePort:            445,
 	SmbdPort:                  445,
 	MetricsExporterMode:       "disabled",
+	ImagePullPolicy:           "IfNotPresent",
 }
 
 // OperatorConfig is a type holding general configuration values.
@@ -73,6 +74,9 @@ type OperatorConfig struct {
 	PodNamespace string `mapstructure:"pod-namespace"`
 	// PodIP is a (string) which defines the currnt pod cluster-ip.
 	PodIP string `mapstructure:"pod-ip"`
+	// ImagePullPolicy is a (string) value which defines the image-download
+	// strategy of samba containers.
+	ImagePullPolicy string `mapstructure:"image-pull-policy"`
 }
 
 // Validate the OperatorConfig returning an error if the config is not
@@ -122,6 +126,7 @@ func NewSource() *Source {
 	v.SetDefault("pod-name", d.PodName)
 	v.SetDefault("pod-namespace", d.PodNamespace)
 	v.SetDefault("pod-ip", d.PodIP)
+	v.SetDefault("image-pull-policy", d.ImagePullPolicy)
 	return &Source{v: v}
 }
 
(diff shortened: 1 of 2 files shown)


synarete:
> In general, I like the idea. Two thoughts though: I'm not sure about the necessity for the case insensitivity. Why not just do a simple string match? Also, I think our default should be the same as the k8s default - if that's ifNotPresent, great. Otherwise, I'd update to match the k8s default. If we want to a make a change for the CI runs, we can do that via the kustomize config files as a CI step.

Why not simple string match: ease of use. ifNotPresent, ifnotpresent, IfNotPresent -- all valid. But as I see that is is confusing, I can fix.

I do not know what is the default value and where is it defined for each different cluster (minikube, OpenShift, k3). The idea is to enforce optimal default for testing. That said, I hoped it would fix our CI timeout failures, but it does _not_. So, I would recommend we keep this one as draft for now and re-consider post `v0.2` release.

phlogistonjohn:
> > In general, I like the idea. Two thoughts though: I'm not sure about the necessity for the case insensitivity. Why not just do a simple string match? Also, I think our default should be the same as the k8s default - if that's ifNotPresent, great. Otherwise, I'd update to match the k8s default. If we want to a make a change for the CI runs, we can do that via the kustomize config files as a CI step.
> 
> Why not simple string match: ease of use. ifNotPresent, ifnotpresent, IfNotPresent -- all valid. But as I see that is is confusing, I can fix.
> 

Got it. I'd just prefer to match k8s behavior with the name. If k8s is case sensitive we should do the same.

> I do not know what is the default value and where is it defined for each different cluster (minikube, OpenShift, k3). The idea is to enforce optimal default for testing. 

OK, maybe we support an additional value to "leave it blank"?  Just a thought, not a requirement.

> That said, I hoped it would fix our CI timeout failures, but it does _not_. So, I would recommend we keep this one as draft for now 

Sure

> and re-consider post `v0.2` release.

That was already what I was thinking. ;-)

phlogistonjohn:
Just a reminder that this PR is still open. If you plan on following up with it, great. If not, or not with the next 3 months, it may be good to close it. It can always be reopened.

synarete:
> Just a reminder that this PR is still open. If you plan on following up with it, great. If not, or not with the next 3 months, it may be good to close it. It can always be reopened.

@phlogistonjohn I rebased this PR over latest master and fixed the above comment. I think now, when we are post `v0.1` it is worth merging this one into master.

internal/conf/config.go
```diff
@@ -22,6 +22,7 @@ var DefaultOperatorConfig = OperatorConfig{
 	SmbServicePort:            445,
 	SmbdPort:                  445,
 	MetricsExporterMode:       "disabled",
+	ImagePullPolicy:           "ifNotPresent",
```
anoopcs9, line 25: As per [docs](https://pkg.go.dev/k8s.io/api@v0.24.3/core/v1#PullPolicy) isn't it correct to use `IfNotPresent` and not `ifNotPresent`?

synarete: @anoopcs9 Good catch! thanks, will fix immediately.

anoopcs9 (approved):
lgtm, thanks.

phlogistonjohn (approved):
LGTM, thanks for fixing this up!

PR #215, opened 2022-06-08
makefile: minor fixes needed to match planned tagging format

phlogistonjohn:
We want the makefile to help us embed the version and hash of the git
checkout used for the build. Fix getting the hash when a tag is applied
and when the tag is in the form vX.Y (rather than vX.Y.Z).


phlogistonjohn:
For expedience I already applied this change for the release, however I figured we can do a proper review at our normal leisurely pace having gotten the release done. :-)

synarete (approved):
LGTM

anoopcs9 (approved):

PR #216, opened 2022-06-14
OpenShift with Metrics

synarete:
Deploy samba-operator over OpenShift cluster, with enabled metrics-exporter side-car container. Takes care of all missing OpenShift configuration by reconcile of SmbCommonConfig.

This patch-set is intended for next development cycle (v0.3)


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.16
```
phlogistonjohn, line 4: Commit msg typo: "scarping"

internal/resources/openshift.go
```diff
@@ -0,0 +1,674 @@
...
+// labels for OpenShift deployment
+func (m *OpenShiftManager) getOrUpdateNamespace(
+	ctx context.Context, nsname string) (*corev1.Namespace, bool, error) {
+	// ---
+	nsCurr, err := m.getNamespace(ctx, nsname)
+	if err != nil {
+		m.logger.Error(err, "Failed to get Namespace", "nsname", nsname)
+		return nil, false, err
+	}
+	labelKey := "openshift.io/cluster-monitoring"
```
phlogistonjohn, line 304: Something I saw in another codebase that I think we should apply: use `const foo = "bar"` even in the function scope  if there's no expectation for the value of foo to be changed.

phlogistonjohn (commented):
I'm partly thru looking at these, dropping a few quick comments now.
And a thought: is there any way we can test *some* of this in the CI?  If not, can we add some tests that are meant to execute on openshift even if we don't have on in our ci yet (and then later add something like microshift to execute it)?

internal/resources/openshift.go
synarete: I agree. Will fix.

internal/resources/openshift.go
```diff
@@ -0,0 +1,675 @@
...
+	}
+}
+
+// Process is called by the controller on reconciliation of
+// SmbCommonConfig
+func (m *OpenShiftManager) Process(
+	ctx context.Context,
+	nsname types.NamespacedName) Result {
+	// Do-nothing if not on OpenShift
+	if !IsOpenShiftCluster(ctx, m.client, m.cfg) {
```
phlogistonjohn, line 69: Stray thought, not required for this PR:  if we see that this gets callsed often enough it may be beneficial to cache the result of this call on the manager instance. We know that it "can't" change, so fetching it from memory ought to be way faster than reading from a k8s api call.

internal/resources/openshift.go
```diff
@@ -0,0 +1,675 @@
...
+	if len(m.cfg.ServiceAccountName) > 0 {
+		return types.NamespacedName{Name: m.cfg.ServiceAccountName}
+	}
+	return types.NamespacedName{
+		Namespace: m.cfg.PodNamespace,
+		Name:      serviceAccountName,
+	}
+}
+
+func (m *OpenShiftManager) getOrCreateSccRole(
```
phlogistonjohn, line 390: This is really trivial but I might as well say it: I think that common idiom for go is to retain all capital letters for initialisms in a camel case name. So instead of "getSccRole" I think "getSCCRole" is considered more correct.  Like I said, minor, but it's something I've been trying to remember to do when I write new code.

internal/resources/openshift.go
```diff
@@ -0,0 +1,675 @@
...
+		return nil, false, err
+	}
+	roleWant := &rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: roleKey.Namespace,
+			Name:      roleKey.Name,
+		},
+		Rules: []rbacv1.PolicyRule{
+			{
+				// TODO: ask John if he prefers import openshift for those vals
```
phlogistonjohn, line 408: I think this is fine for now.  It's certainly no worse than having a bunch of yaml files in the source tree.

internal/resources/openshift.go
```diff
@@ -0,0 +1,675 @@
...
+	cfg *conf.OperatorConfig) bool {
+	key := rtclient.ObjectKey{
+		Namespace: cfg.PodNamespace,
+		Name:      cfg.PodName,
+	}
+	clusterType, err := resolveClusterTypeByPod(ctx, reader, key)
+	return (err == nil) && (clusterType == ClusterTypeOpenshift)
+}
+
+// resolveClusterTypeByPod finds the kind of K8s cluster via anbnotation of one
```
phlogistonjohn, line 652: typo: "anbnotation" -> "annotation"

phlogistonjohn:
Things are looking pretty decent IMO. Some of the openshift-iness goes a little over my head, but I can follow a fair portion of it.
I have a bunch of small non-functional things that need fixing.
We can discuss testing later in a call or in a followup issue.

internal/resources/openshift.go
synarete: I agree; however, we should keep in mind that this in-memory state is ternary: un-initialized, true or false.  At any given moment k8s-scheduler may decide to restarted the samba-operator as a different process on a different node.

internal/resources/openshift.go
synarete: ACK -- will fix.

internal/resources/openshift.go
synarete: Great!

internal/resources/openshift.go
synarete: Thanks -- for some reason vscode spell-checker did not catch this one.

internal/resources/openshift.go
phlogistonjohn: Agree 100%. I generally prefer to define a tristate with a custom type like `type MyTriState int` `const (StateUnknown = MyTriState(0); StateBad = ...; StateGood = ...)`.
Sometimes people like to use a `*bool`. Which is also valid, but I feel it is not as clear/"self documenting".

phlogistonjohn:
I'm not sure if you think it's ready for another round of review, but I note that the "scarping" commit message typo still seems to be present.
Let me know when you think its good to go for reviews and I'll take another look. Thanks!

go.mod
synarete: Missed this one -- will fix

PR #223, opened 2022-06-18
Integration test organizational cleanups

phlogistonjohn:
I started working on making the tests more robust and immediately became annoyed with our approach to organization as well a the semi-non-deterministic execution of the test due to our use of a map.

This may be a bit overkill but I hope it makes things more maintainable and less prone to becoming a jumble if parts are copy-n-pasted in the future.

The new TestGroup type allows a much more consistent and deterministic way of grouping test suites and other (child) test groups. Beyond this PR the hope is that one only needs to think about the test group very rarely and can mostly copy the structure of the `init`functions in the other files that have test suites to execute.

Because determinism could possibly lead to implicit dependencies between tests I also amused myself by adding a Shuffle method that can be invoked through an environment variable. This causes the tests to be run in a randomized order. It is not enabled by default. I have not committed to the idea, but I am considering enabling it for the scheduled runs of the CI (but not PRs). Either way, its there in case we (I) want it.

This is part one of many(?) organizational changes of the integration tests as I hack my way through the weeds that have grown in the test code.

### DIFF
--- a/tests/integration/integration_test.go
+++ b/tests/integration/integration_test.go
@@ -4,8 +4,10 @@
 package integration
 
 import (
+	"math/rand"
 	"sort"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/suite"
 )
@@ -32,6 +34,10 @@ type TestingGroup interface {
 
 type Prioritized interface {
 	Priority() int
+}
+
+type Shuffler interface {
+	Shuffle()
 }
 
 type TestGroup struct {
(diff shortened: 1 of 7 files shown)


tests/integration/integration_test.go
```diff
@@ -4,24 +4,126 @@
...
+
+// ChildPriority returns a new child test group with the provided
+// name and priority value.
+func (tg *TestGroup) ChildPriority(name string, prio int) *TestGroup {
+	child := &TestGroup{name: name, prio: prio}
+	tg.children = append(tg.children, child)
+	sort.SliceStable(tg.children, func(i, j int) bool {
+		pi := 0
+		pj := 0
+		if p, ok := tg.children[i].(Prioritized); ok {
```
synarete, line 80: I find it a bit confusing: could it be that 'ok == false' ? How come? and if yes, do we still get a valid sort or should we panic?

tests/integration/integration_test.go
```diff
@@ -4,24 +4,126 @@
...
+// contents of the subobjects will be shuffled.
+func (tg *TestGroup) Shuffle() {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	for _, c := range tg.children {
+		if schild, ok := c.(Shuffler); ok {
+			schild.Shuffle()
 		}
 	}
+	r.Shuffle(len(tg.children), func(i, j int) {
+		tg.children[i], tg.children[j] = tg.children[j], tg.children[i]
```
synarete, line 107: Oh! is this indeed valid in Go? Does the compiler creates behind-the-scenes the tmp-element which is needed for this swap-assignment? Or is it a potential ordering-bug?

synarete (commented):
Defiantly an improvement compared to previous testing code.

tests/integration/integration_test.go
phlogistonjohn: It's valid and it comes right out of the example for shuffle: https://pkg.go.dev/math/rand#Shuffle  :-)
Here's a stackoverflow I found that goes more in depth about multiple value assignments:
https://stackoverflow.com/questions/48836122/golang-multiple-assignment-evaluation

tests/integration/integration_test.go
phlogistonjohn: The type of the chidlren slice holds `[]TestingGroup` rather than `[]TestGroup`. The TestingGroup interface doesn't require to also be a Prioritized interface. So you could define your own type that meets TestingGroup and isn't Prioritized. 
I had thought about making priority a property of the TestGroup by making a new type like `type childGroup struct { priority int;  g TestingGroup }` and changing the slice to `children []childGroup`.... but I think I just like interfaces so I did this approach which makes priority optional.
If you think priority should be mandatory, we can change it to the other approach I outlined here.

tests/integration/integration_test.go
synarete: Fascinating! I was not aware of those semantics in Go.

tests/integration/integration_test.go
synarete: OK, now I get it; in that case, no need to change it.

synarete (approved):
LGTM

phlogistonjohn:
Thanks for looking at this PR so quickly!

PR #224, opened 2022-06-18
More integration test organizational cleanups

phlogistonjohn:
Depends on: #223

There were far too many context.TODO calls in the integration tests and they were reproducing. This also prepares for adding information about the parameters of the test for logging/tracking purposes.


phlogistonjohn:
I'll keep re-running the centos ci job as needed but I'll point out that the PR that depends on this #225 resolves some of the common ci flakes. I'd appreciate it if the team reviewed this without worrying too much about having the CI pass as I may opt to just ignore centosci until #225.

phlogistonjohn:
@synarete @spuiuk @anoopcs9 - please review. Despite our current CI woes I do think this is an improvement. I am willing to merge this even with some CI failures if we agree the failures are not due to the code change in question while we work through the CI issues.  Thanks!

spuiuk (approved):
Apart from that unused global declared. TestIDKey, eveything thing is fine.

Acked-by: Sachin Prabhu

PR #228, opened 2022-06-29
controllers: cache cluster-type in memory to avoid redundant calls

synarete:
Cache current cluster type within SmbCommonConfigReconciler in order to
avoid extra calls to API server. The in-memory state if set once upon
first call to SmbCommonConfigReconciler.Reconcile.


controllers/smbcommonconfig_controller.go
```diff
@@ -57,12 +58,28 @@ func (r *SmbCommonConfigReconciler) Reconcile(
...
 	res := mgr.Process(ctx, req.NamespacedName)
 	err := res.Err()
 	if res.Requeue() {
 		return ctrl.Result{Requeue: true}, err
 	}
+
+	// Cache in-memory cluster-type to avoid extra network round-trips in next
+	// reconcile phase.
+	if r.ClusterType == "" {
+		r.Log.Info("Keep in-memory", "ClusterType", mgr.ClusterType)
```
phlogistonjohn, line 79: If I was an admin reading this and not a coder I may be confused by this log line. Maybe something like "Saving discovered cluster type" or something along those lines:

synarete: ACK.

phlogistonjohn (approved):
LGTM Thanks

PR #229, opened 2022-07-05
smbcc: Add 'dos charset' option to Global smb.conf defaults

anoopcs9:
`dos charset` option allows us to specify a particular charset which in turn helps to avoid the following warning message on clients in the absence of 850 charset:

> dos charset 'CP850' unavailable - using ASCII


phlogistonjohn (approved):
That was quick! Thanks for taking this one!

phlogistonjohn:
OK, threre's also an open-coded uses of smbclient package in mount_path_test.go that I didn't notice before. I'm filing an issue on that and will follow up.

phlogistonjohn:
Filed #230

Issue #230, opened 2022-07-06
mount_path_test.go makes use of smbclient package without waiting for login


phlogistonjohn:
There are calls to smbclient.CommandOutput & smbclient.Command that have no retry loops like the one added recently for share_access_test.go. We should ensure we can simply login at all before doing more complex uses of the smbclient.

PR #231, opened 2022-07-06
Fix uses of smbclient before waiting for login OK

phlogistonjohn:
Fixes: #230 

Before using the smbclient package to perform smb commands, use the
more robust requireSMBLogin function to test/require that login to
the specified share is possible.


tests/integration/mount_path_test.go
```diff
@@ -92,18 +92,19 @@ func (s *MountPathSuite) SetupSuite() {
 	}
 
 	// Create folders over smbclient
-	smbclient := smbclient.MustPodExec(s.tc, testNamespace,
+	smbClient := smbclient.MustPodExec(s.tc, testNamespace,
```
anoopcs9, line 95: s/smbClient/client ?

I think we used _client_ in recent changes which is  also less confusing.

synarete (approved):
LGTM

tests/integration/mount_path_test.go
phlogistonjohn: Sure.  I often have trouble being consistent with naming. Possibly, because I'm never truly happy by my past choices. ;-)

synarete: smbclient vs smbClient is like potatoes vs potatoes, isn't it? ;)

anoopcs9 (commented):
lgtm, thanks.

anoopcs9 (approved):

PR #232, opened 2022-07-11
integration tests: use wait-loop to check resources cleanup

synarete:
Use simple wait-loop to ensure all k8s resources used by test are no
longer active. In case not all resources have been removed, emit the
relevant details to log.

### DIFF
--- a/tests/integration/cleanup_test.go
+++ b/tests/integration/cleanup_test.go
@@ -133,27 +133,75 @@ func (s *ShareCreateDeleteSuite) getCurrentResources() resourceSnapshot {
 		require = s.Require()
 	)
 
-	rs.pods, err = s.tc.Clientset().CoreV1().
+	rs.pods = &corev1.PodList{}
+	pods, err := s.tc.Clientset().CoreV1().
 		Pods(s.destNamespace).List(ctx, opts)
 	require.NoError(err)
-	rs.services, err = s.tc.Clientset().CoreV1().
+	for _, pod := range pods.Items {
+		if pod.GetDeletionTimestamp().IsZero() {
+			rs.pods.Items = append(rs.pods.Items, pod)
+		}
+	}
+
+	rs.services = &corev1.ServiceList{}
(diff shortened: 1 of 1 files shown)


tests/integration/cleanup_test.go
```diff
@@ -227,18 +227,60 @@ func (s *ShareCreateDeleteSuite) TestCreateAndDelete() {
...
-		len(rs2.deployments.Items), len(existing.deployments.Items))
-	require.Equal(
-		len(rs2.statefulSets.Items), len(existing.statefulSets.Items))
+
+	s.requireResourcesDone(&existing)
+}
+
+func (s *ShareCreateDeleteSuite) requireResourcesDone(base *resourceSnapshot) {
+	allDone := false
+	for i := 0; i < 50 && !allDone; i++ {
```
phlogistonjohn, line 236: Can you please write this in terms of either a [require.Eventually](https://pkg.go.dev/github.com/stretchr/testify/require#Eventually) or our home grown [TryUntil](https://github.com/samba-in-kubernetes/samba-operator/blob/e27ebc6a72b5a92ef74c8c110d283300320b33aa/tests/utils/poll/poll.go#L70) ?

A purely time-based (as opposed to counter) timeout is far easier for a human in the future to reason about and modify IMO.

phlogistonjohn (changes requested):

tests/integration/cleanup_test.go
synarete: Good point -- I should have do so in the first place. Will fix.

phlogistonjohn (dismissed):
LGTM Thanks!  I like the use of logf to help pinpoint what differs now that the require.Equal lines won't be aiding with that.

synarete:
> LGTM Thanks! I like the use of logf to help pinpoint what differs now that the require.Equal lines won't be aiding with that.

Please wait we merge to master -- while testing over OpenShift I realized that we may need another patch on top of this one.

phlogistonjohn:
OK.

PR #233, opened 2022-07-21
support configuration changes

phlogistonjohn:
These changes add the ability to change the configuration of an existing share and have that configuration change reflected in the pod(s) hosting the share.

In addition to the general infrastructure needed to make configuration changes the operator will honor edits to the `readOnly` and `browseable` fields.

Changes to the readOnly field are then used to create a functional test for the feature.

### DIFF
--- a/tests/integration/resource_update_test.go
+++ b/tests/integration/resource_update_test.go
@@ -1,0 +1,265 @@
+//go:build integration
+// +build integration
+// SPDX-License-Identifier: Apache-2.0
+
+package integration
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/stretchr/testify/suite"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	sambaoperatorv1alpha1 "github.com/samba-in-kubernetes/samba-operator/api/v1alpha1"
+	"github.com/samba-in-kubernetes/samba-operator/tests/utils/kube"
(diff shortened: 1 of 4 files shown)


tests/integration/resource_update_test.go
```diff
@@ -0,0 +1,265 @@
...
+type SmbShareUpdateSuite struct {
+	suite.Suite
+
+	commonSources   []kube.FileSource
+	smbShareSources []kube.FileSource
+	shareName       string
+	testAuths       []smbclient.Auth
+	destNamespace   string
+
+	// cached values
```
synarete, line 39: Style: I find those comment redundant. This type is self-explanatory; no need for those extra comments. IMHO a single comment before the struct declaration is more then enough here.

synarete (approved):
LGTM

tests/integration/resource_update_test.go
phlogistonjohn: You're right. I had been doing that as a reminder to myself which variables are meant as "inputs" vs stuff that shouldn't be treated as an input. Because all the types are private I wanted that reminder to myself. However if I want to be serious about that it may be better to use another technique like a newFoo method or even an embedded struct, where the embedded struct is the "input" values.  Regardless of what we do in the future I agree with you that the comment is not hepful and I'll try to remove them in the future.

spuiuk (approved):
Acked-By: Sachin Prabhu <sprabhu@redhat.com>

Do we need to update the documentation?

PR #235, opened 2022-08-05
Break Update reconcile function into smaller parts

phlogistonjohn:
First, we add a Yield method that will return true if the result's content indicates that no further processing should be done.

Then using the Result.Yield method, we can more easily separate parts of the rather large Update function into sub-routines each returning `Done` if that sub-routine has nothing else to do or setting an error or requesting a `Requeue`.

View with -w (ignore whitespace) as well as default to see that most of this patch is moving chunks of Update into new functions.

I think this is a nice small tweak on the existing Result type that enables building "nested" reconcile functions. I'm throwing ths out there as a Draft first to see what you think.

Depends on: #239

### DIFF
--- a/internal/resources/result.go
+++ b/internal/resources/result.go
@@ -31,6 +31,11 @@ func (r Result) Requeue() bool {
 	return r.requeue
 }
 
+// Yield returns true if current processing should be discontinued.
+func (r Result) Yield() bool {
+	return r.requeue || r.err != nil
+}
+
 var (
 	// Done represents a result that is complete.
 	Done = Result{}
--- a/internal/resources/smbshare.go
+++ b/internal/resources/smbshare.go
@@ -131,20 +131,14 @@ func (m *SmbShareManager) Update(
 		return Requeue
 	}
 
-	destNamespace := instance.Namespace
-	cm, created, err := m.getOrCreateConfigMap(ctx, instance, destNamespace)
-	if err != nil {
-		return Result{err: err}
-	} else if created {
-		m.logger.Info("Created config map")
-		return Requeue
-	}
-	planner, changed, err := m.updateConfiguration(ctx, cm, instance)
-	if err != nil {
-		return Result{err: err}
-	} else if changed {
-		m.logger.Info("Updated config map")
-		return Requeue
+	var planner *pln.Planner
+	if p, result := m.updateConfigMap(ctx, instance); !result.Yield() {
+		// p and result only exist within the scope of the if-statement. This
+		// is done to keep the result out of the function scope. But to reuse
+		// the planner, we need to assign p to the func scoped var
+		planner = p
+	} else {
+		return result
 	}
 
 	if shareNeedsPvc(instance) {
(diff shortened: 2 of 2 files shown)


internal/resources/result.go
```diff
@@ -31,6 +31,11 @@ func (r Result) Requeue() bool {
 	return r.requeue
 }
 
+// Yield returns true if current processing should be discontinued.
+func (r Result) Yield() bool {
```
synarete, line 35: Now that I see the code I get you idea. Very nice -- indeed, it makes the code more readable.

internal/resources/smbshare.go
```diff
@@ -148,146 +148,31 @@ func (m *SmbShareManager) Update(
...
-			ctx, statefulSet,
-			int32(planner.SmbShare.Spec.Scaling.MinClusterSize))
-		if err != nil {
-			return Result{err: err}
-		} else if resized {
-			m.logger.Info("Resized statefulSet")
-			return Requeue
+		result := m.updateClusteredState(ctx, planner)
+		if result.Yield() {
+			return result
```
synarete, line 164: Code-style consistency: why not have `if result := ... ; result.Yield() {` (within same line), as done in other places?

internal/resources/smbshare.go
```diff
@@ -148,146 +148,31 @@ func (m *SmbShareManager) Update(
...
-
-		resized, err := m.updateDeploymentSize(ctx, deployment)
-		if err != nil {
-			return Result{err: err}
-		} else if resized {
-			m.logger.Info("Resized deployment")
-			return Requeue
+		result := m.updateNonClusteredState(ctx, planner)
+		if result.Yield() {
+			return result
```
synarete, line 169: Code-style consistency: why not have `if result := ... ; result.Yield() {` (within same line), as done in other places?

internal/resources/smbshare.go
```diff
@@ -322,6 +207,189 @@ func (m *SmbShareManager) Update(
...
+	// recorded backend to check if a change is being made after the
+	// "point of no return".
+	// In the future we should certainly revisit this and support
+	// intelligent methods to handle changes like this.
+	b := instance.Annotations[serverBackend]
+	if planner.IsClustered() && b != clusteredBackend {
+		err = fmt.Errorf(
+			"Can not convert SmbShare to clustered instance."+
+				" Current backend: %s",
+			b)
```
synarete, line 284: Style: I find this line-break confusing.

internal/resources/smbshare.go
```diff
@@ -322,6 +207,189 @@ func (m *SmbShareManager) Update(
...
+			"SmbShare.Namespace", instance.Namespace,
+			"SmbShare.Name", instance.Name,
+			"SmbShare.UID", instance.UID)
+		return Result{err: err}
+	}
+	if !planner.IsClustered() && b != standardBackend {
+		err = fmt.Errorf(
+			"Can not convert SmbShare to non-clustered instance."+
+				" Current backend: %s",
+			b)
```
synarete, line 297: Style: I find this line-break confusing.

synarete:
Overall, looks very nice!
Few minor style comments.

internal/resources/smbshare.go
phlogistonjohn: Good point. Originally the function took more parameters and the line was very long. I can change this back to the `if x := f(); x {` style now that the lines are no longer extra long.

internal/resources/smbshare.go
phlogistonjohn: OK. It was like that in the previous version of the code before it was moved (you can see this if you use "ignore whitespace" feature).  If you want I can add a follow up patch or if its not a blocker for this PR, we can leave it and the next time we work on the code we can improve the formatting.

internal/resources/smbshare.go
phlogistonjohn: Fixed.

synarete (approved):
LGTM

phlogistonjohn:
centos ci is stuck? I'm going to poke it again but right now it appears to be still pending since yesterday (?):  ` Pending — Build triggered for merge commit. `

anoopcs9:
> `Pending — Build triggered for merge commit.`

This is the current status for **centos-ci/sink-clustered/mini-k8s-1.23** and was a side effect of some changes done at jenkins deployment. We should be looking at **centos-ci/sink-clustered/mini-k8s-1.24**.

phlogistonjohn:
Oh, I totally missed that. Thanks.

phlogistonjohn:
Since that 1.23 item probably isn't going to get unstuck I'm joint going to merge with admin power.

internal/resources/result.go
```diff
@@ -31,6 +31,11 @@ func (r Result) Requeue() bool {
...
+	return r.requeue || r.err != nil
```
spuiuk, line 36: Should we use result.Requeue() here instead?

phlogistonjohn: Requeue returns the value of requeue. It's a public accessor method for that field. So this direct use of the value is just "cutting out the middle man".  If Requeue() became fancy in the future and was the product of multiple inputs we could simply upate Yield as well.

phlogistonjohn: update Yield as well.... to use Requeue() then.

PR #239, opened 2022-08-10
Add comments to some packages

phlogistonjohn:
This will fix errors detected by revive as part of running `make check`.
It's genuinely a good thing to have docs, but in some cases they could use a little fleshing out.

### DIFF
--- a/controllers/smbshare_controller.go
+++ b/controllers/smbshare_controller.go
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package controllers defines this operator's controllers.
 package controllers
 
 import (
--- a/internal/conf/config.go
+++ b/internal/conf/config.go
@@ -1,3 +1,4 @@
+// Package conf defines the operator's configuration parameters.
 package conf
 
 import (
--- a/internal/planner/doc.go
+++ b/internal/planner/doc.go
@@ -1,0 +1,2 @@
+// Package planner helps determine the desired state.
+package planner
--- a/internal/resources/doc.go
+++ b/internal/resources/doc.go
@@ -1,0 +1,2 @@
+// Package resources works directly with k8s resources.
+package resources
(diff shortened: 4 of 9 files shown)


anoopcs9:
I see that in some places package comments were added via new _doc.go_ and for others existing files were used. Is there a particular reason to do so?

synarete:
> I see that in some places package comments were added via new _doc.go_ and for others existing files were used. Is there a particular reason to do so?

As far as I can tell, not all places have `doc.go`. Thus, if `doc.go` exists, place the package comment there; otherwise, use another major file.

synarete (approved):
LGTM.
Tested with go1.19 on fc35.x86_64

anoopcs9:
> > I see that in some places package comments were added via new _doc.go_ and for others existing files were used. Is there a particular reason to do so?
> 
> As far as I can tell, not all places have `doc.go`. Thus, if `doc.go` exists, place the package comment there; otherwise, use another major file.

Yes, but the patch adds **new** _doc.go_ files under _internal/planner_, _internal/resources_, _internal/smbcc_ and _tests/utils/kube_ whereas comments could have been included within existing files under those directories/packages.

Checking again now I think the criteria is more like minimum number of source files present under a particular directory/package(>3) qualifying for an extra _doc.go_ 😆. But then `package smbcc` should not have one 🤨

phlogistonjohn:
> I see that in some places package comments were added via new _doc.go_ and for others existing files were used. Is there a particular reason to do so?

Yeah, if the the package was a single file and I expect it to stay that way for the near term I just added a comment to the existing file. If the package had many files I added a doc.go (doc.go is a common convention in Go).

phlogistonjohn:
> 
> Checking again now I think the criteria is more like minimum number of source files present under a particular directory/package(>3) qualifying for an extra _doc.go_ laughing. But then `package smbcc` should not have one raised_eyebrow

I might break that one up in the future.

anoopcs9 (approved):
> > Checking again now I think the criteria is more like minimum number of source files present under a particular directory/package(>3) qualifying for an extra _doc.go_ laughing. But then `package smbcc` should not have one raised_eyebrow
> 
> I might break that one up in the future.

Fine by me.

PR #240, opened 2022-08-17
grouped shares: hosting multiple shares in one server instance

phlogistonjohn:
For brevity I'll be calling this feature "grouped shares" from now on. What this does is allow one to create multiple SmbShare resources that will all be hosted by the same server group instance. For your standard deployment based instances this means a single smbd will host multiple shares; multiple shares will be accessible at a single "hostname" (smb://foo.ns.cluster.test/abc & smb://foo.ns.cluster.test/xyz).

The goal of this feature is largely one of resource efficiency. The ability to have SmbShares fully isolated from each other is a powerful default, allowing for things like mixing and matching AD domains or user-level security on a single logical unit (k8s cluster). However, having a large number of SmbShare resources can leave to a fair bit of overhead as each SmbShare will end up creating a pod running multiple daemons. As long as the shares are compatible - they share key settings like securityConfig and pvc, etc - we will permit grouping those SmbShares under one "instance".

This feature is enabled by adding the scaling parameters "Group" and "GroupMode". GroupMode must be set to "explicit" and a name provided at Group. The group name must be a valid k8s resource name.
Example:
```yaml
apiVersion: samba-operator.samba.org/v1alpha1
kind: SmbShare
metadata:
  name: share1
spec:
  shareName: "Share One"
  readOnly: false
  browseable: true
  securityConfig: sharesec1
  scaling:
    groupMode: explicit
    group: grouped1
    # ^^^ this is the new stuff
  storage:
    pvc:
      name: mypvc
      path: one
```

This finally implements one of the major planned features from the [phase 1 design](https://github.com/samba-in-kubernetes/samba-operator/blob/master/docs/design/crd-proposal-phase1.md) that had yet to be implemented.

One small downside of requiring grouped shares to be configured explicitly is that the creator of the SmbShares must plan ahead if the shares should be grouped or not. I think this is a small downside as I envion this feature mainly being used by tools that automatically provision shares. Imagine a CSI-alike that creates shares en-masse. That's largely the use case I'm thinking of for this. Of course, nothing stops a human admin from making use of it.

IMPORTANT: I am not planning on implementing and testing grouped shares for clustered instances in this PR. I was originally going to - but in order to get this PR reviewed earlier rather than later I have decided to work on combining the two scaling features later. I think that since clustered instances are experimental, this is an acceptable approach. 

Also out of scope is the ability to change from ungrouped to grouped. Setting group mode to "never" (or leaving it unset) currently really does mean never. Similarly, it's not supported to change "explicit" to "never". Perhaps in the future we'll make it possible to migrate from one mode to another.


internal/resources/smbshare.go
```diff
@@ -196,6 +198,13 @@ func (m *SmbShareManager) updateConfigMap(
 		m.logger.Info("Created config map")
 		return nil, Requeue
 	}
+	changed, err := m.claimOwnership(ctx, smbshare, cm)
+	if err != nil {
+		return nil, Result{err: err}
+	} else if changed {
```
synarete, line 204: Prefer `if changed` (no need to have `else if`).

internal/resources/smbshare.go
```diff
@@ -339,6 +348,14 @@ func (m *SmbShareManager) updateClusteredState(
 		return Requeue
 	}
 
+	changed, err := m.claimOwnership(ctx, planner.SmbShare, statefulSet)
+	if err != nil {
+		return Result{err: err}
+	} else if changed {
```
synarete, line 354: Prefer `if changed` (no need to have `else if`).

internal/resources/smbshare.go
```diff
@@ -370,6 +387,14 @@ func (m *SmbShareManager) updateNonClusteredState(
 		return Requeue
 	}
 
+	changed, err := m.claimOwnership(ctx, planner.SmbShare, deployment)
+	if err != nil {
+		return Result{err: err}
+	} else if changed {
```
synarete, line 393: Prefer `if changed` (no need to have `else if`).

internal/resources/smbshare.go
```diff
@@ -384,17 +409,27 @@ func (m *SmbShareManager) updateSmbService(
...
 		return Result{err: err}
 	} else if created {
 		m.logger.Info("Created service")
 		return Requeue
 	}
+
+	changed, err := m.claimOwnership(ctx, planner.SmbShare, svc)
+	if err != nil {
+		return Result{err: err}
+	} else if changed {
```
synarete, line 424: Prefer `if changed` (no need to have `else if`).

synarete (commented):
Only few minor comments. 
I want to test it over OpenShift (make sure we did not miss something) before final approval.

synarete (approved):
Tested on OpenShift. Looks good.

LGTM

internal/resources/events.go
```diff
@@ -27,4 +27,7 @@ const (
 	ReasonCreatedPersistentVolumeClaim = "CreatedPersistentVolumeClaim"
 	ReasonCreatedDeployment            = "CreatedDeployment"
 	ReasonCreatedStatefulSet           = "CreatedStatefulSet"
+	// ReasonInvalidConfiguration indicates a failure to proceed due
+	// to an invalid configuration.
```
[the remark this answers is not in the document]
phlogistonjohn: I was going to reply that I agree that the wording of the comment was a bit redundant but we needed it because it's an exported type. But when I looked at the file I see that the other constants don't have one. So I'm not entirely sure why I thought it was needed but the other commands don't.  I'll look into it.

internal/resources/ownership.go
```diff
@@ -0,0 +1,86 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+
+	sambaoperatorv1alpha1 "github.com/samba-in-kubernetes/samba-operator/api/v1alpha1"
+)
+
+func smbShareOwnerRefs(obj metav1.Object) ([]metav1.OwnerReference, error) {
+	found := []metav1.OwnerReference{}
+	ssgvk := sambaoperatorv1alpha1.GroupVersion
```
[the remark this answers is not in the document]
phlogistonjohn: >  Curiosity-only: what does ssgvk stands for?
Smb Share Group Version Kind.
> IMHO smbgv would be a better variable name.
Probably. :-)

internal/resources/smbshare.go
phlogistonjohn: There are a lot of:
```
if err != nil {
   return ...
} else if abc {
   return ...
}
```

pattern in the current smbshare.go. I am going to leave them alone for now but we can do a follow up pr that just does a style change to *all* of these in one batch. Sound ok?

internal/resources/ownership.go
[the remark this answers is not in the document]
phlogistonjohn: Fixed.

internal/resources/events.go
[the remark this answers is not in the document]
phlogistonjohn: Fixed.

internal/resources/smbshare.go
synarete: Sure. I will take a closer look on this pattern cross all code-base.

phlogistonjohn: Resolving this conversation as we've discussed we'll do a general if-else-if with returns style clean-up in a follow up PR.

PR #241, opened 2022-08-21
dockerfile: use ubi9 as base image

synarete:
Use Red Hat's Universal Base Image 9 (ubi9) for samba-operator.


phlogistonjohn (approved):
LGTM

PR #244, opened 2022-09-12
controllers: removed unused Scheme field

synarete:
Having an unused and uninitialized 'Scheme' member to SmbCommonConfigReconciler is confusing; a naive user may assume that it is set to non-nil reference which end with run-time panic.

### DIFF
--- a/controllers/smbcommonconfig_controller.go
+++ b/controllers/smbcommonconfig_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 
 	"github.com/go-logr/logr"
-	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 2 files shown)


phlogistonjohn (dismissed):
Good find, thanks.

phlogistonjohn (approved):
LGTM

anoopcs9 (commented):
Don't we have similar situation with `SmbSecurityConfigReconciler`?

synarete:
> Don't we have similar situation with `SmbSecurityConfigReconciler`?

@anoopcs9 Good catch! I will update this PR with appropriate fix.

phlogistonjohn:
... and centos ci passed too :exclamation: :exclamation: :exclamation:

anoopcs9:
> ... and centos ci passed too exclamation exclamation exclamation

🕺🏻 🕺🏻

synarete:
> > ... and centos ci passed too exclamation exclamation exclamation
> 
> 🕺🏻 🕺🏻

Cool!

PR #246, opened 2022-09-13
hack: update to golangci-lint v1.49.0

synarete:
Use newer version of golangci-lint. Disabled linters which are defined as deprecated or unmaintained (namely: deadcode, structcheck and varcheck).

Fixed newly discovered trivial case of assigned-but-unused variable in configuration.go.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -5,7 +5,7 @@ linters:
   disable-all: true
   enable:
     - asciicheck
-    - deadcode
+    # - deadcode
     - depguard
     - dupl
     - errcheck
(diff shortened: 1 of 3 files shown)


synarete:
golangci-lint v1.49 requires go1.19.

synarete:
> > golangci-lint v1.49 requires go1.19.
> 
> That's unfortunate. I really don't like it when projects stop supporting versions of Go before the Go team does. Can we take a release of golangci-lint a release or two back that fixes our typecheck errors but also works with Go 1.18?

Sure, I will try it.

synarete:
> > > golangci-lint v1.49 requires go1.19.
> > 
> > 
> > That's unfortunate. I really don't like it when projects stop supporting versions of Go before the Go team does. Can we take a release of golangci-lint a release or two back that fixes our typecheck errors but also works with Go 1.18?
> 
> Sure, I will try it.

@phlogistonjohn I dived into it an realized that its a more complicated issue: currently, samba-operator supports as far back as Go1.16 (e.g., in github actions). However, not all our build tools are compatible with this version in their latest release. Thus, we need to pin-point each tool to explicit version which is not too old but still supports Go1.16. A separate PR is on its way.

PR #247, opened 2022-09-13
planner: fixed set-but-not-used assignment

synarete:
Fixed minor dead-assigment to use un-named variable. Detected by golangci-lint v1.49, which requires go1.19.


anoopcs9 (approved):
lgtm.

anoopcs9 (approved):

PR #248, opened 2022-09-14
explicit tools versions

synarete:
Currently, samba-operator supports as far back as Go1.16 (e.g., in github actions). However, not all our build tools are compatible with this version in their latest release. Thus, we need to pin-point each tool to explicit version which is not too old but still supports Go1.16.

### DIFF
--- a/hack/install-tools.sh
+++ b/hack/install-tools.sh
@@ -33,7 +33,7 @@ _install_golangci_lint() {
 }
 
 _install_yq() {
-	_install_tool github.com/mikefarah/yq/v4@latest
+	_install_tool github.com/mikefarah/yq/v4@v4.23.1
 }
 
 case "$1" in
--- a/hack/install-tools.sh
+++ b/hack/install-tools.sh
@@ -29,7 +29,7 @@ _install_revive() {
 }
 
 _install_golangci_lint() {
-	_install_tool github.com/golangci/golangci-lint/cmd/golangci-lint@v1.43.0
+	_install_tool github.com/golangci/golangci-lint/cmd/golangci-lint@v1.45.2
 }
 
 _install_yq() {
--- a/hack/install-tools.sh
+++ b/hack/install-tools.sh
@@ -25,7 +25,7 @@ _install_controller_gen() {
 }
 
 _install_revive() {
-	_install_tool github.com/mgechev/revive@latest
+	_install_tool github.com/mgechev/revive@v1.2.3
 }
 
 _install_golangci_lint() {


phlogistonjohn:
> Currently, samba-operator supports as far back as Go1.16 (e.g., in github actions). However, not all our build tools are compatible with this version in their latest release. Thus, we need to pin-point each tool to explicit version which is not too old but still supports Go1.16.

This is somewhat of a historical accident. In my mind the policy is not "support Go 1.16" it is "support all the current versions of Go". Therefore in a number of cases we can't say "latest" because that would only build/test with the newest version of Go and we run the risk of using stuff that won't work on Go. If there was a symbolic name in the setup-go action, I'd probably use it... but I can't find such a thing so we're stuck bumping up the version numbers from time to time.

Sometimes, when I'm letting my distribution lag and I'm using an older version on my dev machine(s) I don't mind the CI lagging along with me. However, that's out of laziness and I can usually be talked out of that without much effort. :-)

All this said, I think this change is a fine start and there's nothing wrong with this!

phlogistonjohn (dismissed):

phlogistonjohn:
@synarete @anoopcs9 how do we want to handle the centos ci failures for a change like this. I'm fine with merging it w/o a centos ci as it's fairly small and mainly directed at the github checks, which passed. Any thoughts?

synarete:
> @synarete @anoopcs9 how do we want to handle the centos ci failures for a change like this. I'm fine with merging it w/o a passing centos ci run as it's fairly small and mainly directed at the github checks, which passed. Any thoughts?

As far as I can tell from CI log, those failures are not related to the actual changes in the PR. @anoopcs9 Could you please take a look as well? (just to be on the safe side).

anoopcs9:
> > @synarete @anoopcs9 how do we want to handle the centos ci failures for a change like this. I'm fine with merging it w/o a passing centos ci run as it's fairly small and mainly directed at the github checks, which passed. Any thoughts?
> 
> As far as I can tell from CI log, those failures are not related to the actual changes in the PR. @anoopcs9 Could you please take a look as well? (just to be on the safe side).

As informed via other channels, _TestIntegration/smbSharesClustered/withDNS/TestShareAccessByDomainName_ started failing for every run after recent migration to reserve nodes for testing from new EC2 pool in CentOS CI infra. This is currently under investigation #252

phlogistonjohn:
I appreciate the update. Getting back to my original question: regardless of the cause and severity of the current centos ci issues, does anyone mind just merging this without waiting for the ci issues to be resolved? I think it's small scoped enough to do so, personally.

anoopcs9 (approved):

anoopcs9 (changes requested):
Can we have the following typos fixed in commit messages?

* _smabe-operator_ in https://github.com/samba-in-kubernetes/samba-operator/pull/248/commits/ab8a22d68bc95fe10304a984e83f988bdd3faa69
* _smae_ in https://github.com/samba-in-kubernetes/samba-operator/pull/248/commits/52055f1e92a06141e737f8eb4ecd234c3cddf622

synarete:
> Can we have the following typos fixed in commit messages?
> 
> * _smabe-operator_ in [ab8a22d](https://github.com/samba-in-kubernetes/samba-operator/commit/ab8a22d68bc95fe10304a984e83f988bdd3faa69)
> * _smae_ in [52055f1](https://github.com/samba-in-kubernetes/samba-operator/commit/52055f1e92a06141e737f8eb4ecd234c3cddf622)

Sure, will fix

phlogistonjohn (approved):

anoopcs9 (approved):
lgtm, thanks.

PR #249, opened 2022-09-14
planner: fixes and tests for configuration update

phlogistonjohn:
There are a number of code-paths the  planner takes to update the smbcc
configuration. This initial set of unit tests gets us to around 77%
coverage of configuration.go, checking some of the most common of these
paths.

### DIFF
--- a/internal/planner/configuration.go
+++ b/internal/planner/configuration.go
@@ -101,7 +101,7 @@ func (pl *Planner) Update() (changed bool, err error) {
 	}
 	cfgKey := pl.instanceID()
 	cfg, found := pl.ConfigState.Configs[cfgKey]
-	if !found || cfg.Shares[0] != shareKey {
+	if !found {
 		cfg = smbcc.ConfigSection{
 			Shares:       []smbcc.Key{shareKey},
 			Globals:      []smbcc.Key{smbcc.Globals},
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
Note that updating the configuration with a 2nd share is not supported by the operator today. The fix here will be needed when we start supporting that as a valid action in the operator. That said, it's nice to have the library function work correctly anyways :-)

internal/planner/configuration.go
```diff
@@ -81,11 +81,11 @@ func (pl *Planner) idmapOptions() smbcc.SmbOptions {
 // Update the held configuration based on the state of the instance
 // configuration.
 func (pl *Planner) Update() (changed bool, err error) {
-	globals, found := pl.ConfigState.Globals[smbcc.Globals]
+	_, found := pl.ConfigState.Globals[smbcc.Globals]
 	if !found {
 		globalOptions := smbcc.NewGlobalOptions()
 		globalOptions.SmbPort = pl.GlobalConfig.SmbdPort
-		globals = smbcc.NewGlobals(globalOptions)
+		globals := smbcc.NewGlobals(globalOptions)
```
anoopcs9, line 88: This is interesting..

Above change was already merged as part of #247 and GitHub couldn't find any conflict for the current PR?

phlogistonjohn: I just rebased. git didn't prompt me to resolve conflicts either, so my guess is that both cli git and github determined it was not a conflict that needed manual resolution. Regardless, the rebase has now eliminated that difference.

internal/planner/configuration_test.go
```diff
@@ -0,0 +1,192 @@
...
+	}, state)
+
+	// apply changes
+	changed, err := p.Update()
+	assert.NoError(t, err)
+	assert.True(t, changed)
+
+	assert.Len(t, state.Shares, 1)
+	assert.Len(t, state.Configs, 1)
+	assert.Len(t, state.Globals, 2)
```
anoopcs9, line 188: I had to go through the _Update()_ func to realize that `Realm` from `SmbSecurityConfigSpec` was the one getting added as the next(2nd) key in `Globals` map.

anoopcs9 (approved):
Change looks OK to me.

phlogistonjohn:
Giving this until Thursday (09-28) for any additional reviews.

PR #250, opened 2022-09-14
controllers: set controller watches on owned configmaps & statefulsets

phlogistonjohn:
When a configmap or statefulset is created by the operator in response to an SmbShare resource, we set the SmbShare as the owner (controller owner). We want our reconcile loop to be called when a change is made to one of these resources.

### DIFF
--- a/controllers/smbshare_controller.go
+++ b/controllers/smbshare_controller.go
@@ -83,6 +83,9 @@ func (r *SmbShareReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&sambaoperatorv1alpha1.SmbShare{}).
 		Owns(&corev1.PersistentVolumeClaim{}).
+		Owns(&corev1.ConfigMap{}).
+		Owns(&corev1.Service{}).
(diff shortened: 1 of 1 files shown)


controllers/smbshare_controller.go
```diff
@@ -83,6 +83,8 @@ func (r *SmbShareReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&sambaoperatorv1alpha1.SmbShare{}).
 		Owns(&corev1.PersistentVolumeClaim{}).
+		Owns(&corev1.ConfigMap{}).
 		Owns(&appsv1.Deployment{}).
+		Owns(&appsv1.StatefulSet{}).
```
anoopcs9, line 88: I don't remember the discussion outcome but what about `corev1.Service`?

phlogistonjohn: Sure, anything resource type w generate from an SmbShare is fair game (although not required).

anoopcs9 (approved):
lgtm, thanks.

synarete (approved):
LGTM

PR #251, opened 2022-09-14
resources: fix config update container creating missing share dirs

phlogistonjohn:
The configuration update container will create missing share directories if they don't exist (as well as setting up initial perms). However, this will not work if the container doesn't have access to the data volume. Give the container access to the data volume.

### DIFF
--- a/internal/resources/pods.go
+++ b/internal/resources/pods.go
@@ -427,10 +427,13 @@ func buildSmbdCtrs(
 	ctrs := []corev1.Container{}
 	ctrs = append(ctrs, buildSmbdCtr(planner, env, vols))
 	metaOnlyVols := vols.exclude(tagData)
-	// insert a container to watch for changes to the config json
-	// and apply those changes to samba
+	// Insert a container to watch for changes to the config json
+	// and apply those changes to samba.
+	// We need to pass data vols to the config update container as it is
+	// responsible for creating missing dirs and setting perms when new shares
+	// appear.
 	ctrs = append(ctrs, buildUpdateConfigWatchCtr(
-		planner, env, metaOnlyVols))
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
I 100% hit the merge button and github *closed* the PR on me. Two times! This one I carefully made sure I hit the correct button.... sure enough. blammo. closed pr!

internal/resources/pods.go
```diff
@@ -427,10 +427,13 @@ func buildSmbdCtrs(
 	ctrs := []corev1.Container{}
 	ctrs = append(ctrs, buildSmbdCtr(planner, env, vols))
 	metaOnlyVols := vols.exclude(tagData)
-	// insert a container to watch for changes to the config json
-	// and apply those changes to samba
+	// Insert a container to watch for changes to the config json
+	// and apply those changes to samba.
+	// We need to pass data vols to the config update container as it is
+	// responsible for creating missing dirs and setting perms when new shares
+	// appear.
```
anoopcs9, line 434: I have to admit that I am not yet 100% confident with the current flow of actions from various containers. But still does that mean it is a bug? Or is it like a fix for future when configuration gets updated for additional shares?

phlogistonjohn: Mostly the latter. I have a very large WIP series of patches in #240 where I originally found the problem. The problem does exist in the current version in the case where sambacc is supposed to change permissions, but it's such a rare case that it is hard to hit. With the changes in the new PR it's very obvious because it fails to add new share subdirs to the PVC.

anoopcs9: Thanks for the clarification.

synarete (approved):
LGTM

PR #243, opened 2022-08-31
tests/integration: Fix issues reported by `gofmt`

anoopcs9:
Following diff was suggested as a result of `gofmt` execution:

```
# go version
go version go1.18.4 linux/amd64

# gofmt -d -e -s -l .
tests/integration/path_permissions.go
diff -u tests/integration/path_permissions.go.orig tests/integration/path_permissions.go
--- tests/integration/path_permissions.go.orig	2022-08-31 21:40:37.504171615 +0530
+++ tests/integration/path_permissions.go	2022-08-31 21:40:37.504171615 +0530
@@ -1,3 +1,4 @@
+//go:build integration
 // +build integration
 
 package integration
tests/integration/resource_update_test.go
diff -u tests/integration/resource_update_test.go.orig tests/integration/resource_update_test.go
--- tests/integration/resource_update_test.go.orig	2022-08-31 21:40:37.507171634 +0530
+++ tests/integration/resource_update_test.go	2022-08-31 21:40:37.507171634 +0530
@@ -1,5 +1,6 @@
 //go:build integration
 // +build integration
+
 // SPDX-License-Identifier: Apache-2.0
 
 package integration
```

PR #253, opened 2022-09-20
Go version >= 1.17

synarete:
Set minimal supported version of Go to 1.17. Updated github workflow accordingly.

Prefer 1.17 over 1.18 as some of our auxiliary build tools (linters) are still buggy when used with go1.18 features (namely, generics).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v2
         with:
-          go-version: "1.16"
+          go-version: "1.17"
       - uses: actions/checkout@v2
       - name: Build
         run: make
@@ -23,7 +23,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v2
         with:
-          go-version: "1.16"
+          go-version: "1.17"
       - uses: actions/checkout@v2
       - name: Install revive
         run: go install github.com/mgechev/revive@latest
@@ -34,7 +34,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v2
         with:
-          go-version: "1.16"
+          go-version: "1.17"
       - uses: actions/checkout@v2
       - name: run the tests
         run: make test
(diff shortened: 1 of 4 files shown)


synarete:
> Everything looks good with one minor caveat - the list of packages in go.mod looks longer that I expected (I expected new stuff in that 2nd section, but not that much). Can you please verify you ran `go mod tidy`? If so, I'll happily approve - I just want to double check the list isn't longer than necessary.

I re-run `go mod tidy` and got same result, both with `go1.17` and `go1.18`. May need to do deeper dive into https://go.dev/ref/mod#go-mod-file-require but I would prefer to do so in a separate PR.

phlogistonjohn (approved):
LGTM

phlogistonjohn:
I'll wait approx. one day before merging for other feedback. I will ignore the centos ci status for this PR and merge w/o waiting for success from that CI.

tests/integration/path_permissions.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration
```
anoopcs9, line 1: #243 includes this change.

tests/integration/resource_update_test.go
```diff
@@ -1,5 +1,6 @@
 //go:build integration
 // +build integration
+
```
anoopcs9, line 3: #243 includes this change.

internal/planner/configuration.go
```diff
@@ -81,11 +81,11 @@ func (pl *Planner) idmapOptions() smbcc.SmbOptions {
 // Update the held configuration based on the state of the instance
 // configuration.
 func (pl *Planner) Update() (changed bool, err error) {
-	globals, found := pl.ConfigState.Globals[smbcc.Globals]
+	_, found := pl.ConfigState.Globals[smbcc.Globals]
 	if !found {
 		globalOptions := smbcc.NewGlobalOptions()
 		globalOptions.SmbPort = pl.GlobalConfig.SmbdPort
-		globals = smbcc.NewGlobals(globalOptions)
+		globals := smbcc.NewGlobals(globalOptions)
```
anoopcs9, line 88: This is same as #247 , right?

anoopcs9 (changes requested):
Few changes from this PR already exists as separate PRs(see below comments). I prefer to consider those older PRs and accordingly modify the current one.

phlogistonjohn (approved):
LGTM, thanks.

anoopcs9 (approved):
lgtm.

PR #255, opened 2022-09-22
Lint code using gosec 

synarete:
Inspects source code for security problems using [gosec](https://github.com/securego/gosec)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -206,9 +206,9 @@ bundle-build:
 	@false
 	# See vcs history for how this could be restored or adapted in the future.
 
-.PHONY: check check-revive check-golangci-lint check-format check-yaml
+.PHONY: check check-revive check-golangci-lint check-format check-yaml check-gosec
 
-check: check-revive check-golangci-lint check-format vet check-yaml
+check: check-revive check-golangci-lint check-format vet check-yaml check-gosec
 
 check-format:
 	! $(GOFMT_CMD) $(CHECK_GOFMT_FLAGS) . | sed 's,^,formatting error: ,' | grep 'go$$'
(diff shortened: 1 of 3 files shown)


internal/resources/volumes.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	userSecretVolName = "users-config"
+	userSecretVolName = "users-config" // #nosec
```
phlogistonjohn, line 27: I'm very curious, what what did the tool think was wrong with this line?

synarete: @phlogistonjohn  As noted in the corresponding rule [g101](https://securego.io/docs/rules/g101.html), "this can be noisy and may generate false positives" (which is indeed, the case here).

phlogistonjohn: Oh, it picks up on the term "Secret" that's pretty annoying. But thanks for explaining it.

phlogistonjohn (approved):
LGTM

internal/resources/volumes.go
anoopcs9, line 27: Shouldn't we just disable rule [G101](https://securego.io/docs/rules/g101.html) as mentioned [here](https://github.com/securego/gosec#annotating-code)?

```suggestion
	userSecretVolName = "users-config" //#nosec G101
```

anoopcs9 (commented):
Looks fine..just a clarification needed. See below:

internal/resources/volumes.go
phlogistonjohn: Come to think of it if it is such a noisy rule, can we just disable G101 on the command line?

phlogistonjohn: To partially answer my own question, after looking at the gosec readme: something like `gosec -exclude=G101 ...` should work.

synarete: ACK

anoopcs9 (approved):
lgtm, thanks.

phlogistonjohn (approved):
LGTM, thanks!

PR #258, opened 2022-09-29
planner: add Prune function to remove shares

phlogistonjohn:
Nothing uses this change yet, but it adds a function to the internal planner
library to Prune shares from an otherwise populated smbcc config.
Once the operator supports adding multiple shares to one instance we'll also
need a way to remove them.

With unit tests.

### DIFF
--- a/internal/planner/configuration.go
+++ b/internal/planner/configuration.go
@@ -117,7 +117,7 @@ func (pl *Planner) Update() (changed bool, err error) {
 		}
 		changed = true
 	}
-	if !hasShare(cfg, shareKey) {
+	if !hasShare(&cfg, shareKey) {
 		cfg.Shares = append(cfg.Shares, shareKey)
 		changed = true
 	}
@@ -143,6 +143,24 @@ func (pl *Planner) Update() (changed bool, err error) {
 			}
 			changed = true
 		}
+	}
+	return
+}
+
+// Prune the target share from the configuration.
+func (pl *Planner) Prune() (changed bool, err error) {
+	cfgKey := pl.instanceID()
+	shareKey := smbcc.Key(pl.shareName())
+
+	if cfg, found := pl.ConfigState.Configs[cfgKey]; found {
+		if removeShare(&cfg, shareKey) {
+			pl.ConfigState.Configs[cfgKey] = cfg
+			changed = true
+		}
+	}
+	if _, found := pl.ConfigState.Shares[shareKey]; found {
+		delete(pl.ConfigState.Shares, shareKey)
+		changed = true
 	}
 	return
 }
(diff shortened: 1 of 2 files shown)


internal/planner/configuration.go
```diff
@@ -147,6 +147,24 @@ func (pl *Planner) Update() (changed bool, err error) {
 	return
 }
 
+// Prune the target share from the configuration.
+func (pl *Planner) Prune() (changed bool, err error) {
```
synarete, line 151: I find it a bit confusing that the function returns `err error` but it is not defined anywhere within the function's body. But maybe it is just me still needs to adjust my eyes to Go style.

internal/planner/configuration.go
```diff
@@ -181,3 +199,18 @@ func hasShare(cfg smbcc.ConfigSection, k smbcc.Key) bool {
 	}
 	return false
 }
+
+func removeShare(cfg *smbcc.ConfigSection, k smbcc.Key) bool {
```
synarete, line 203: You are using here pass-by-pointer `cfg *smbcc.ConfigSection`; for consistency, you should do so also for `hasShare` above (kind of slipped my eyes in previous review).

internal/planner/configuration.go
```diff
@@ -181,3 +199,18 @@ func hasShare(cfg smbcc.ConfigSection, k smbcc.Key) bool {
...
+		if cfg.Shares[i] == k {
+			idx = i
+			break
+		}
+	}
+	if idx == -1 {
+		return false
+	}
+	cfg.Shares = append(cfg.Shares[:idx], cfg.Shares[idx+1:]...)
+	return true
```
synarete, line 215: I believe the Go community prefers the following code style:

```
	for i := range cfg.Shares {
		if cfg.Shares[i] == k {
			cfg.Shares = append(cfg.Shares[:i], cfg.Shares[i+1:]...)
			return true
		}
	}
	return false
```

phlogistonjohn: I had it written that way originally and ran into a weird issue, but it may have been something else I had mistakenly done.  Since I have unit tests for this now  I can try changing it back and seeing what happens.

internal/planner/configuration.go
phlogistonjohn: Usually, when possible I like to pass by value on purpose to have unshared data. The hasShare function is "read only" whereas removeShare mutates the config section. So it's "intentionally" inconsitent. In this case it's totally fine to change it for consistency, but there may be cases where I push back more.  I'll try to note those in comments in the future.

internal/planner/configuration.go
phlogistonjohn: I often don't use that style much, but every once in a while I think it's clearer.  In this case I'm not married to it.

spuiuk (approved):
Ack

phlogistonjohn:
I would pay money for some kind of feedback that I've issued the correct command to rerun the jenkins tests.

internal/planner/configuration.go
synarete: ACK

internal/planner/configuration.go
synarete: ACK

synarete (approved):
LGTM

PR #260, opened 2022-09-29
planner: add functions to check if smbshares are compatible 

phlogistonjohn:
Compatible SmbShares are shares that can be hosted by one smbd instance.
In other words, compatible share can be grouped together. This function
is not yet used but will be once we have the ability to specify grouped
smbshares.

Adds a newer version of testify and matching unit tests.

### DIFF
--- a/internal/planner/compatible.go
+++ b/internal/planner/compatible.go
@@ -1,0 +1,102 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package planner
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// IncompatibleInstanceError indicates SmbShare resources are not compatible
+// with each other and can not hosted by the same server.
+type IncompatibleInstanceError struct {
+	current  string
+	existing string
+	reason   string
+}
+
+// Error interface method.
+func (iie IncompatibleInstanceError) Error() string {
+	return fmt.Sprintf("Share resource %s is incompatible with %s: %s",
+		iie.current,
+		iie.existing,
+		iie.reason)
+}
+
+func incompatible(
+	current, existing InstanceConfiguration,
+	reason string) IncompatibleInstanceError {
+	// ---
+	return IncompatibleInstanceError{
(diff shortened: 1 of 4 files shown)


internal/planner/compatible.go
```diff
@@ -0,0 +1,102 @@
...
+// Error interface method.
+func (iie IncompatibleInstanceError) Error() string {
+	return fmt.Sprintf("Share resource %s is incompatible with %s: %s",
+		iie.current,
+		iie.existing,
+		iie.reason)
+}
+
+func incompatible(
+	current, existing InstanceConfiguration,
```
synarete, line 28: I prefer using pass-by-pointer in such cases. Although `InstanceConfiguration` is just a 4-pointers tuple now, it may get bigger in the future causing inefficient pass-by-value.

internal/planner/compatible.go
```diff
@@ -0,0 +1,102 @@
...
+		current:  current.SmbShare.Name,
+		existing: existing.SmbShare.Name,
+		reason:   reason,
+	}
+}
+
+// CheckCompatible returns an error if the instance configurations are
+// not compatible with each other. A compatible instance is one that
+// can share the same smbd.
+func CheckCompatible(current, existing InstanceConfiguration) error {
```
synarete, line 41: Same as previous comment: prefer pass-by-pointer.

synarete (approved):
LGTM. 
Minor comment on pass-by-value vs pass-by-pointer.

anoopcs9 (approved):
lgtm

internal/planner/compatible.go
phlogistonjohn: This is a case where I personally feel semantics > performance UNTIL we profile and discover this is leading to a measurable performance issue. I want the "read only" nature of this comparison to be apparent.

On top of that https://codewithyury.com/golang-pass-by-pointer-vs-pass-by-value/ notes that pass by value is not always slower. I one read a longer and more thorough article saying the same but I can't find it right now.

PR #263, opened 2022-10-19
workflows: use checkout@v3 and setup-go@v3

synarete:
When using github action checkout we get an annoying annotation warning "Node.js 12 actions are deprecated". Resolved by upgrading to v3.

See also: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


anoopcs9 (approved):
lgtm.

PR #264, opened 2022-10-27
resources: fixed confusing if-else-if patterns

synarete:
The code fragment:
```
    if cond1 {
	return foo
    } else if cond2 {
	do-bar
    }
```
may be replaced with:
```
    if cond1 {
	return foo
    }
    if cond2 {
	do-bar
    }
```
The second pattern is considered more readable (also in other programming languages[1]).

[1] https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html

### DIFF
--- a/internal/resources/smbshare.go
+++ b/internal/resources/smbshare.go
@@ -116,7 +116,8 @@ func (m *SmbShareManager) Update(
 	changed, err := m.addFinalizer(ctx, instance)
 	if err != nil {
 		return Result{err: err}
-	} else if changed {
+	}
+	if changed {
 		m.logger.Info("Added finalizer")
 		return Requeue
 	}
@@ -128,7 +129,8 @@ func (m *SmbShareManager) Update(
 	changed, err = m.setServerGroup(ctx, instance)
 	if err != nil {
 		return Result{err: err}
-	} else if changed {
+	}
+	if changed {
 		m.logger.Info("Updated server group")
 		return Requeue
 	}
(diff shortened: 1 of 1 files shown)


internal/resources/smbshare.go
```diff
@@ -116,7 +116,8 @@ func (m *SmbShareManager) Update(
 	changed, err := m.addFinalizer(ctx, instance)
 	if err != nil {
 		return Result{err: err}
-	} else if changed {
+	}
+	if changed {
 		m.logger.Info("Added finalizer")
 		return Requeue
```
obnoxxx, line 122: couldn't use of`defer return Requeue` or similar simplify this case even more?

obnoxxx (commented):
Quite a few commits have beenn added to this PR that do not relate to its original topic of improving if-else patterns at all but rather relate to grouping of shares. why is that? 

commits in a PR should imho implement what the PR description states. if ta PR depends on another unmerged PR, it should be mentioned explicitly.

Also, regarding commit messages and PR titles, I suggest to use present tense (or imperative) rather than past tense. e.g. "Fix confusing if-else-patterns " instead of "fixed confusing if-else-patterns.

phlogistonjohn:
Yes, this PR is based on the patches from #240 which just got merged yesterday. I think @synarete has not gotten a chance to rebase it yet. :-)

obnoxxx (approved):
Thanks for rebasing and for adjusting the commit message. LGTM nbow!

PR #271, opened 2022-12-27
Style code blocks as YAML in `howto.md`

yardenshoham:
They were set as plain text, but these are YAML code blocks


phlogistonjohn (approved):
LGTM. Thanks very much for these helpful doc improvements!

PR #275, opened 2023-01-27
github:add mergify configuration

obnoxxx:
taken from samba-container / sambacc and adjusted.

Signed-off-by: Michael Adam <obnox@samba.org>


.github/mergify.yml
```diff
@@ -0,0 +1,80 @@
...
+  # Perform automatic merge on conditions
+  - name: Automatic merge on approval
+    conditions:
+      - check-success=check
+      - check-success=build
+      - check-success=podmanbuild
+      - check-success=dockerbuild
+      - check-success=test
+      - check-success=test-kubernetes
+      -check-success=centos-ci/sink-clustered/mini-k8s-1.25
```
obnoxxx, line 46: not quite sure yet if this will work. it deviates from the pattern of the other checks ...

phlogistonjohn: The name appears to match, so I think it'll be all right. If not we can just fix it later.

phlogistonjohn:
In the *commit message* there should be a space after the colon.

phlogistonjohn:
Also, in case you missed it there are syntax errors in the YAML that mergify is failing on.

.github/mergify.yml
phlogistonjohn: One of the syntax errors is on this line. There should be a space after the first dash.

anoopcs9: > not quite sure yet if this will work. it deviates from the pattern of the other checks ...

This should work. As John mentioned above a space(after hyphen `-`) is all that is missing from the above line.

.github/mergify.yml
```diff
@@ -0,0 +1,80 @@
...
+    conditions:
+      - check-success=check
+        - check-success=build
+      - check-success=podmanbuild
+      - check-success=dockerbuild
+      - check-success=test
+      - check-success=test-kubernetes
+      -check-success=centos-ci/sink-clustered/mini-k8s-1.25
+
+        - check-success=dpulls
```
anoopcs9, line 25: List of checks are not aligned correctly and space after initial hyphen(`-`) is still not present. See `-check-success=centos-ci/sink-clustered/mini-k8s-1.25`

.github/mergify.yml
```diff
@@ -0,0 +1,80 @@
...
+  # Perform automatic merge on conditions
+  - name: Automatic merge on approval
+    conditions:
+      - check-success=check
+      - check-success=build
+      - check-success=podmanbuild
+      - check-success=dockerbuild
+      - check-success=test
+      - check-success=test-kubernetes
+      -check-success=centos-ci/sink-clustered/mini-k8s-1.25
```
anoopcs9, line 46: Same as above.
Space is missing after initial hyphen(`-`)

anoopcs9 (changes requested):

.github/mergify.yml
obnoxxx: fixed, thanks!

.github/mergify.yml
obnoxxx: fixed, thanks!

obnoxxx:
@anoopcs9  - now the whole `make check` and mrgify's config check both pass! :-)

.github/mergify.yml
```diff
@@ -0,0 +1,80 @@
...
+  - name: Automatic merge on approval
+    conditions:
+      - check-success=check
+      - check-success=build
+      - check-success=podmanbuild
+      - check-success=dockerbuild
+      - check-success=test
+      - check-success=test-kubernetes
+      - check-success=centos-ci/sink-clustered/mini-k8s-1.25
+
```
anoopcs9, line 47:
```
- check-success=dpulls
```

is missing here.

.github/mergify.yml
```diff
@@ -0,0 +1,80 @@
...
+  - name: default
+    conditions:
+      - check-success=check
+      - check-success=build
+      - check-success=podmanbuild
+      - check-success=dockerbuild
+      - check-success=test
+      - check-success=test-kubernetes
+      - check-success=centos-ci/sink-clustered/mini-k8s-1.25
+
```
anoopcs9, line 24: Empty new line??

.github/mergify.yml
```diff
@@ -0,0 +1,80 @@
...
+      # suffixed by their username) in order to indicate a detailed review has
+      # been started and not completed. This will hold the PR until the
+      # label has been removed.
+      - "-label~=^review-in-progress"
+      - "base=master"
+      # Even if there are 2 or more approvals we won't automerge if there are
+      # any changes requested.
+      - "#changes-requested-reviews-by=0"
+      - or:
+      # Any contributor's PR can be automerged with 2 (or more) reviews.
```
anoopcs9, line 62: Following warning should go away if we indent the comment properly.

anoopcs9 (changes requested):

.github/mergify.yml
obnoxxx: > Following warning should go away if we indent the comment properly.
It is  gone now, @anoopcs9 . the checks have all passed in the CI.

phlogistonjohn: It's a warning not an error, so the CI shows green but it's still showing up if you view the patch on github.

anoopcs9: Correct. It's a warning and thus test passes. While we are at it let's just fix it now.

.github/mergify.yml
anoopcs9: This is still missing in the latest version.

anoopcs9 (changes requested):

.github/mergify.yml
phlogistonjohn: I agree with @anoopcs9. Please fix this when you have a moment

.github/mergify.yml
obnoxxx: > 

added now

obnoxxx: strange, I added it  and see it locally, but not in the github view ... 🤷

.github/mergify.yml
obnoxxx: > 



> I agree with @anoopcs9. Please fix this when you have a moment

Trying to. and I think the commentindentation errors or warnings are gone now, but I see the following locally from yamllint and in the editor:

```
56:6      error    syntax error: expected <block end>, but found '<block sequence start>' (syntax)
  57:7      error    wrong indentation: expected 7 but found 6  (indentation)
```

trying to fix all of this now ...

obnoxxx: I don't understand the block sequence error abd how to fix it. the rest should be taken care of now ...

.github/mergify.yml
```diff
@@ -0,0 +1,80 @@
...
+  - "#changes-requested-reviews-by=0"
+  - or:
+      # Any contributor's PR can be automerged with 2 (or more) reviews.
+      - "#approved-reviews-by>=2"
+      # A maintainer's contribution only needs 1 review BUT we give a grace
+      # period over just two weeks for a 2nd reviewer to hopefully appear.
+  - and:
+      - "updated-at<15 days ago"
+    - "author=@maintainers"
+     - "#approved-reviews-by>=1"
```
anoopcs9, line 69: Above failure is due to wrong alignment from LINENO: 56 through LINENO:69.

anoopcs9 (changes requested):

.github/mergify.yml
obnoxxx: > Above failure is due to wrong alignment from LINENO: 56 through LINENO:69.

@anoopcs9  - not sure what you are referring to with "above", but I updated once more and it looks better now. Still seeing some obtuse failures locally though ...

obnoxxx:
strange, after the latest update, check-yaml fails in the ci, but I can't see the error message in the output in the github UI ... @anoopcs9 , @phlogistonjohn  - any idea?

phlogistonjohn (approved):
LGTM.  @anoopcs9 please take another look at this updated version

.github/mergify.yml
```diff
@@ -0,0 +1,79 @@
...
+          - "#approved-reviews-by>=2"
+          # A maintainer's contribution only needs 1 review BUT we give a grace
+          # period over just two weeks for a 2nd reviewer to hopefully appear.
+      - and:
+          - "updated-at<15 days ago"
+          - "author=@maintainers"
+          - "#approved-reviews-by>=1"
+    actions:
+      queue: {}
+      dismiss_reviews: {}
```
synarete, line 71: At first, I though we may need an explicit `queue.method` (something like [this](https://github.com/mozilla/application-services/blob/main/.mergify.yml#L48-L58)), but if I read the merigfy [queue docs](https://docs.mergify.com/actions/queue/) correctly, then rebase is done automatically. Am I correct?

synarete (approved):
LGTM

.github/mergify.yml
anoopcs9: I think `defaults` section at the very beginning plays a role here in determining the queue method. 
```
The defaults section is used to define default configuration valued for actions run by pull request
rules and by Commands. If the options are defined in  pull_request_rules they are used, otherwise,
the values set in defaults are used.
```

anoopcs9 (approved):

PR #276, opened 2023-02-02
workflows: build within emulated container

synarete:
Use QEMU to build samba-operator within containerized environment; either AMD64 or ARM64 architectures (regardless of local host architecture). Using this method we can ensures that all build tools are properly configured on architectures other than default x86_64, without relaying on specific hardware.

### DIFF
--- a/.github/workflows/build-within-container.yml
+++ b/.github/workflows/build-within-container.yml
@@ -1,0 +1,26 @@
+name: Build within container
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+jobs:
+  build-within-container:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Enable emulation
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y qemu qemu-user-static
+          sudo update-binfmts --display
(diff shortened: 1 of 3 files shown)


synarete:
@phlogistonjohn Fascinating findings! I was able to have same results on my local machine as well (+ more).

1) As far as I remember, I also tried this path with podman, but at the time failed to have those results. It could be due to wrong build args but I suspect that now I using a newer version of podman.

2) I also tried to follow the instructions at https://podman.io/blogs/2021/10/11/multiarch.html but hit some issues so decided to abandon this path. Again, could be related to older podman version.

3) Note that `/proc/cpuinfo` shows local host machine (unlike in the qemu version). However, not a problem in out use case.

4) Following the above results, I suggest we to reject this PR and I will prepare a new one using podman and existing Makefile (with the appropriate changes).

phlogistonjohn:
> [...SNIP...]
>
> 4\. Following the above results, I suggest we to reject this PR and I will prepare a new one using podman and existing Makefile (with the appropriate changes).

Fine by me, PR numbers are not a precious resource.  :-)

Let me know if you plan on doing the new PR or you'd like me to start, as I now have the needed bits to try it out.

phlogistonjohn:
Closing per previous comment.

Issue #283, opened 2023-02-15
How to configure node selector if using a mixed K8S


RonnyRen:
Hi All,
Our Kubernetes cluster is mixed with Linux nodes and Windows nodes, After deploying samba-operator successfully, I created a SMBShare, the new deployment was running on a Windows node and error happened. I have to change the node selector to a Linux node and then it is running, I also need to update PVC if default storage class uses local-path provisioner. I didn't find anything related to this in the document, how can I run it in mixed cluster? Could you please give some suggestion? Thanks in advance.

Thanks,
Ronny

phlogistonjohn:
Thank you for the report. Unfortunately, that is not something we've implemented yet. 

We should probably default our deplyments/etc to always request linux nodes. I also have plans to add more advanced node selection/affinity/etc rules to the SmbCommonConfig.

I'll try to look into this soon, thanks!

phlogistonjohn:
@RonnyRen  I just pushed a new pr - #291 - to implement this using settings in our SmbCommonConfig CR. Even if you don't read Go, the last patch adds a doc update. Feel free to take a look and let us know if this meets your use case. Thanks!

PR #286, opened 2023-02-21
add default node selector value and configuration

phlogistonjohn:
Related To: #283

Adds a `default-node-selector` configuration value and some hard coded default values. The hard coded defaults match what samba-container currently provides, images that work on x86_64 linux. Putting a json  mapping in the config allows an admin/dev to customize the defaults without hacking the code or rebuilding the operator (say for example - experimenting with custom build images for arm64).

I plan on following this up with node selectors and affinity support as part of SmbCommonConfig.

### DIFF
--- a/internal/conf/config.go
+++ b/internal/conf/config.go
@@ -24,6 +24,7 @@ var DefaultOperatorConfig = OperatorConfig{
 	SmbdPort:                  445,
 	MetricsExporterMode:       "disabled",
 	ImagePullPolicy:           "IfNotPresent",
+	DefaultNodeSelector:       "",
 }
 
 // OperatorConfig is a type holding general configuration values.
@@ -78,6 +79,10 @@ type OperatorConfig struct {
 	// ImagePullPolicy is a (string) value which defines the image-download
 	// strategy of samba containers.
 	ImagePullPolicy string `mapstructure:"image-pull-policy"`
+	// DefaultNodeSelector is a string value containing JSON which defines
+	// a set of key-value pairs that will be used for all default node
+	// selection. If left blank, internal defaults will be used.
+	DefaultNodeSelector string `mapstructure:"default-node-selector"`
 }
 
 // Validate the OperatorConfig returning an error if the config is not
@@ -128,6 +133,7 @@ func NewSource() *Source {
 	v.SetDefault("pod-namespace", d.PodNamespace)
 	v.SetDefault("pod-ip", d.PodIP)
 	v.SetDefault("image-pull-policy", d.ImagePullPolicy)
+	v.SetDefault("default-node-selector", d.DefaultNodeSelector)
 	return &Source{v: v}
 }
 
--- a/internal/planner/selector.go
+++ b/internal/planner/selector.go
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package planner
+
+import (
+	"encoding/json"
+)
+
+// builtinDefaultNodeSelector defines our defaults if no administrative
+// overrides are supplied. Samba containers run only on linux. Currently
+// they're only built for x86_64 (amd64).
+var builtinDefaultNodeSelector = map[string]string{
+	"kubernetes.io/os":   "linux",
+	"kubernetes.io/arch": "amd64",
+}
+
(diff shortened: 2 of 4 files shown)


internal/conf/config.go
```diff
@@ -24,6 +24,7 @@ var DefaultOperatorConfig = OperatorConfig{
 	SmbdPort:                  445,
 	MetricsExporterMode:       "disabled",
 	ImagePullPolicy:           "IfNotPresent",
+	DefaultNodeSelector:       "",
```
synarete, line 27: We would probably need to add in the future validation layer to `OperatorConfig` member. It is better to detect illegal inputs (in this case, malformatted json) early on at boot time instead of upon first usage.

internal/planner/selector.go
```diff
@@ -0,0 +1,29 @@
...
+package planner
+
+import (
+	"encoding/json"
+)
+
+// BuiltinDefaultNodeSelector defines our defaults if no administrative
+// overrides are supplied. Samba containers run only on linux. Currently
+// they're only built for x86_64 (amd64).
+var BuiltinDefaultNodeSelector = map[string]string{
```
synarete, line 12: Could be private var (rename to `builtinDefaultNodeSelector` or just `defaultNodeSelector`).

internal/resources/pods.go
```diff
@@ -113,6 +113,9 @@ func buildADPodSpec(
 		buildMustJoinCtr(planner, joinEnv, joinVols),
 	}
 	podSpec.Containers = containers
+	// we have no logger to log the json syntax error to. have to ignore it for now
+	nsel, _ := planner.NodeSelector()
+	podSpec.NodeSelector = nsel
```
synarete, line 118: Consider: `podSpec.NodeSelector, _ = planner.NodeSelector()`

synarete (approved):
LGTM

internal/conf/config.go
phlogistonjohn: Yeah, that's tricky because I also want the config layer to be fairly generic and not have intimate detailed knowledge of what the layers above want.

Here's a thought: what about calling into the planner function somewhere in the main.go file and checking the error there and then blocking startup at that layer?

internal/planner/selector.go
phlogistonjohn: Sure.

internal/resources/pods.go
phlogistonjohn: Sure.

internal/conf/config.go
synarete: Sounds good (in a separate PR)

phlogistonjohn: Ohhh well... take a look, I just pushed it as an additional patch to this one.

internal/planner/selector.go
phlogistonjohn: Resolved.

internal/conf/config.go
synarete: LGTM

phlogistonjohn:
@synarete you'll need to use the approval at the review level if it looks OK because GH already dismissed your earlier LGTM. Thanks!

synarete (approved):

main.go
```diff
@@ -97,6 +98,15 @@ func main() {
...
 
+	planner := pln.New(pln.InstanceConfiguration{
+		GlobalConfig: conf.Get(),
+	}, nil)
+	if _, err := planner.NodeSelector(); err != nil {
+		setupLog.Error(err, "invalid node selector configuration value",
+			"note", "value must be a JSON object containing strings")
+		os.Exit(1)
+	}
+
```
anoopcs9, line 109: Should we place the above validation block before LINENO: 99 such that we say "loaded configuration successfully" after checking `DefaultNodeSelector` too?

phlogistonjohn: The first one indicates that it successfully got a configuration. This new one is saying a value in the config is no good. Let's wait on it and change it if we get users who complain.

anoopcs9: Ok.

anoopcs9 (commented):
lgtm, thanks.

anoopcs9 (approved):

PR #287, opened 2023-02-23
build: Bump minimum Go version to 1.18

anoopcs9:
_go.sum_ changes are result of running `go mod tidy` after updating **_go_** directive in _go.mod_.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
-          go-version: "1.17"
+          go-version: "1.18"
       - uses: actions/checkout@v3
       - name: Build
         run: make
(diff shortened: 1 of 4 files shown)


synarete (approved):
LGTM

.github/workflows/main.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
-          go-version: "1.17"
+          go-version: "1.18"
```
phlogistonjohn, line 16: IIRC I think setup-go now has symbolic names for the current and current-1 go releases. Once we catch up with go we can maybe switch to the symbolic name.

phlogistonjohn (approved):
LGTM

.github/workflows/main.yml
phlogistonjohn: i just looked. It's  called "oldstable"  (and  "stable" for latest stable)

PR #293, opened 2023-03-09
Deploy on OpenShift with dedicated SCC

synarete:
Refactor deployment over OpenShift cluster using custom `samba` SCC. User is expected to deploy the `samba` SCC prior to the deployment of the samba-operator (or, via kustomization mechanism). 

Reconcile ServiceAccount, Role and RoleBinding as part of SmbShare reconcile loop, within the namespace in which the smbshare resides.

### DIFF
--- a/internal/conf/config.go
+++ b/internal/conf/config.go
@@ -9,6 +9,13 @@ import (
 	"github.com/spf13/viper"
 )
 
+const (
+	// ClusterTypeDefault defines the default value for cluster type
+	ClusterTypeDefault = "default"
+	// ClusterTypeOpenShift defines the type-name for OpenShift clusters
+	ClusterTypeOpenShift = "openshift"
+)
+
 // DefaultOperatorConfig holds the default values of OperatorConfig.
 var DefaultOperatorConfig = OperatorConfig{
 	SmbdContainerImage:        "quay.io/samba.org/samba-server:latest",
@@ -25,6 +32,7 @@ var DefaultOperatorConfig = OperatorConfig{
 	MetricsExporterMode:       "disabled",
 	ImagePullPolicy:           "IfNotPresent",
 	DefaultNodeSelector:       "",
+	ClusterType:               "",
 }
 
 // OperatorConfig is a type holding general configuration values.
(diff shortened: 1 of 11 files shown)


internal/conf/config.go
```diff
@@ -9,6 +9,13 @@ import (
 	"github.com/spf13/viper"
 )
 
+const (
+	// ClusterTypeDefault defines the default value for cluster type
+	ClusterTypeDefault = "default"
+	// ClusterTypeOpenshift defines the type-name for OpenShift clusters
+	ClusterTypeOpenshift = "openshift"
```
phlogistonjohn, line 16: This is super-duper nitpicky, and I clearly didn't notice it last time, but shouldn't it be spelled `ClusterTypeOpenShift`?

phlogistonjohn:
I made a few small comments on my first skim. I will def. need to re-read this for a proper review. Please ping me in a few work days if you don't see a full review by then.

internal/resources/pods.go
```diff
@@ -838,3 +841,10 @@ func imagePullPolicy(pl *pln.Planner) corev1.PullPolicy {
 	}
 	return pullPolicy
 }
+
+func ctrPrivSecurityContext() *corev1.SecurityContext {
+	return &corev1.SecurityContext{
+		Privileged:   &[]bool{true}[0],
+		RunAsNonRoot: &[]bool{false}[0],
```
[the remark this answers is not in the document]
synarete: https://sdk.operatorframework.io/docs/best-practices/pod-security-standards/

internal/conf/config.go
synarete: ACK. Will fix

synarete:
> Either in this PR or in a follow up I think it would be good to have a new document in the docs dir that goes over some of the important differences between plain k8s and openshift, I think the most important would be to explain that: if you are using openshift and want to create smbshares in namespaces other than the default (which we've already updated via kustomize) the creator of the ns also needs to set the openshift specific security properties.
> 
> (I'm assuming I got the summary right, feel free to correct me)

I am working on samba-on-openshift doc. Will come in a separate PR.

phlogistonjohn:
I want to try microshift later this week and if it goes OK then I will test this PR on it. If I can't get microshift running either I will probably just approve, but if this goes unreviewed past next tues (03-15) feel free to ping me and remind me I owe you a proper review.

phlogistonjohn (approved):
I built and ran this while following the docs at PR #295. As I now have looked the code over and now have run it, I can strongly say LGTM.

spuiuk (approved):
Ack

PR #298, opened 2023-04-07
Revisit running gitlint in the CI

phlogistonjohn:
After having to fight with actions/checkout in another repo for other reasons I learned a few things that may help us run the commit checks in the github ci

### DIFF
--- a/hack/install-tools.sh
+++ b/hack/install-tools.sh
@@ -5,15 +5,29 @@
 #   GOBIN=<dir> install-tools.sh --<tool-name>
 #
 set -e
-GO_CMD=${GO_CMD:-$(command -v go)}
-PY_CMD="${PY_CMD:-$(command -v python3)}"
-GOBIN=${GOBIN:-${GOPATH}/bin}
 
 _require_gobin() {
 	mkdir -p "${GOBIN}"
 }
 
+_require_go() {
+	if [ -z "$GO_CMD" ]; then
+		echo "error: go command required, but not found" >&2
+		echo "(set GO_CMD to specify go command)" >&2
+		exit 5
+	fi
+}
+
+_require_py() {
+	if [ -z "$PY_CMD" ]; then
+		echo "error: python3 command required, but not found" >&2
+		echo "(set PY_CMD to specify python command)" >&2
+		exit 5
+	fi
+}
+
 _install_tool() {
+	_require_go
 	GOBIN="${GOBIN}" ${GO_CMD} install "$1"
 }
 
@@ -43,10 +57,24 @@ _install_gosec() {
 
 _install_gitlint() {
 	_require_gobin
+	_require_py
 	"${PY_CMD}" -m venv "${GOBIN}/.py"
 	"${GOBIN}/.py/bin/pip" install "gitlint==0.19.1"
 	ln -s "${GOBIN}/.py/bin/gitlint"  "${GOBIN}/gitlint"
 }
+
+GOBIN="${GOBIN:-${GOPATH}/bin}"
+
+if [ -z "$GO_CMD" ]; then
+    if ! GO_CMD="$(command -v go)"; then
+        echo "warning: failed to find go command" >&2
+    fi
+fi
+if [ -z "$PY_CMD" ]; then
+    if ! PY_CMD="$(command -v python3)"; then
+        echo "warning: failed to find python3 command" >&2
+    fi
+fi
 
 case "$1" in
 	--kustomize)
--- a/hack/install-tools.sh
+++ b/hack/install-tools.sh
@@ -6,6 +6,7 @@
 #
 set -e
 GO_CMD=${GO_CMD:-$(command -v go)}
+PY_CMD="${PY_CMD:-$(command -v python3)}"
 GOBIN=${GOBIN:-${GOPATH}/bin}
 
 _require_gobin() {
@@ -40,6 +41,13 @@ _install_gosec() {
 	_install_tool github.com/securego/gosec/v2/cmd/gosec@v2.13.1
 }
 
+_install_gitlint() {
+	_require_gobin
+	"${PY_CMD}" -m venv "${GOBIN}/.py"
+	"${GOBIN}/.py/bin/pip" install "gitlint==0.19.1"
+	ln -s "${GOBIN}/.py/bin/gitlint"  "${GOBIN}/gitlint"
+}
+
 case "$1" in
 	--kustomize)
 		_require_gobin
@@ -65,15 +73,19 @@ case "$1" in
 		_require_gobin
 		_install_gosec
 		;;
+	--gitlint)
+		_install_gitlint
+		;;
 	*)
 		echo "usage: GOBIN=<dir> $0 --<tool-name>"
 		echo ""
-		echo "availabel tools:"
+		echo "available tools:"
 		echo "  --kustomize"
 		echo "  --controller-gen"
 		echo "  --revive"
 		echo "  --golangci-lint"
 		echo "  --yq"
 		echo "  --gosec"
+		echo "  --gitlint"
 		;;
 esac
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,6 +29,18 @@ jobs:
         run: go install github.com/mgechev/revive@latest
       - name: Run checks
         run: make check
+  check-commits:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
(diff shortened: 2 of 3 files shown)


phlogistonjohn:
@synarete what do you think? I think the big differences between this and #274 is that I added gitlint to be auto installed like many of the other tools and then the thing that "makes it work" in the CI is to use the `ref` argument to checkout action as well as using the `origin/` prefix on the branch name.

synarete:
> @synarete what do you think? I think the big differences between this and #274 is that I added gitlint to be auto installed like many of the other tools and then the thing that "makes it work" in the CI is to use the `ref` argument to checkout action as well as using the `origin/` prefix on the branch name.

@phlogistonjohn Installing gitlint locally (like we do with other bin tools) looks great. That said, `make check-gitlint` still fails when integrated with github actions. See: https://github.com/synarete/samba-operator/actions/runs/4649123528/jobs/8227255345

phlogistonjohn:
> > @synarete what do you think? I think the big differences between this and #274 is that I added gitlint to be auto installed like many of the other tools and then the thing that "makes it work" in the CI is to use the `ref` argument to checkout action as well as using the `origin/` prefix on the branch name.
> 
> @phlogistonjohn Installing gitlint locally (like we do with other bin tools) looks great. That said, `make check-gitlint` still fails when integrated with github actions. See: https://github.com/synarete/samba-operator/actions/runs/4649123528/jobs/8227255345

It's possible I overlooked something but I don't think the version you are testing is setting the `ref` parameter for actions/checkout@v3.  I copy and paste the text below from the run you linked:

```
Run actions/checkout@v3
  with:
    fetch-depth: 0
    repository: synarete/samba-operator
    token: ***
    ssh-strict: true
    persist-credentials: true
    clean: true
    lfs: false
    submodules: false
    set-safe-directory: true
Syncing repository: synarete/samba-operator
```

Try setting ` ref: ${{ github.event.pull_request.head.sha }}`  as a sibling to `fetch-depth: 0`, and see if that improves things for you. Notice how in my run it sets ref and it only logs an error for the "bad news" commit - as that is an intential failing commit I use to check the check and I plan to remove before merging

phlogistonjohn:
On second thought, I think I understand now. You're basing this run off of my commits but you are not running them as part of a PR. So while I think you have the key specified in the job, the value `${{ github.event.pull_request.head.sha }}` maps to nothing because you are running it on a "raw" branch not a PR. So there's no pull_request sub section!  We probably don't need this check on anything not a PR anyway so we can add `if: github.event_name == 'pull_request'` to the job so that it only runs if pull_request is going to have data in it.

synarete:
> > > @synarete what do you think? I think the big differences between this and #274 is that I added gitlint to be auto installed like many of the other tools and then the thing that "makes it work" in the CI is to use the `ref` argument to checkout action as well as using the `origin/` prefix on the branch name.
> > 
> > 
> > @phlogistonjohn Installing gitlint locally (like we do with other bin tools) looks great. That said, `make check-gitlint` still fails when integrated with github actions. See: https://github.com/synarete/samba-operator/actions/runs/4649123528/jobs/8227255345
> 
> It's possible I overlooked something but I don't think the version you are testing is setting the `ref` parameter for actions/checkout@v3. I copy and paste the text below from the run you linked:
> 
> ```
> Run actions/checkout@v3
>   with:
>     fetch-depth: 0
>     repository: synarete/samba-operator
>     token: ***
>     ssh-strict: true
>     persist-credentials: true
>     clean: true
>     lfs: false
>     submodules: false
>     set-safe-directory: true
> Syncing repository: synarete/samba-operator
> ```
> 
> Try setting ` ref: ${{ github.event.pull_request.head.sha }}` as a sibling to `fetch-depth: 0`, and see if that improves things for you. Notice how in my run it sets ref and it only logs an error for the "bad news" commit - as that is an intential failing commit I use to check the check and I plan to remove before merging

@phlogistonjohn `ref` is properly set, see https://github.com/synarete/samba-operator/blob/ss-gitlint-review/.github/workflows/main.yml#L32-L42

Yor made a test with `expected-to-fail` commit, while I tried with `should-not-fail` commits. This is the same problem I encountered in my previous iteration with gitlint: when run via github actions `make check-gitlint` failed on valid commits (but does not happen on local repository).

synarete:
> On second thought, I think I understand now. You're basing this run off of my commits but you are not running them as part of a PR. So while I think you have the key specified in the job, the value `${{ github.event.pull_request.head.sha }}` maps to nothing because you are running it on a "raw" branch not a PR. So there's no pull_request sub section! We probably don't need this check on anything not a PR anyway so we can add `if: github.event_name == 'pull_request'` to the job so that it only runs if pull_request is going to have data in it.

@phlogistonjohn Good catch!

synarete (commented):
LGTM

hack/install-tools.sh
```diff
@@ -40,6 +41,13 @@ _install_gosec() {
 	_install_tool github.com/securego/gosec/v2/cmd/gosec@v2.13.1
 }
 
+_install_gitlint() {
+	_require_gobin
```
anoopcs9, line 45: nit: __require_gobin_ was called outside __install_ functions in all other places.

anoopcs9:
> We probably don't need this check on anything not a PR anyway so we can add `if: github.event_name == 'pull_request'` to the job so that it only runs if pull_request is going to have data in it.

What happened to this part? Don't we need to differentiate _pull_request_  and other events?

phlogistonjohn:
> > We probably don't need this check on anything not a PR anyway so we can add `if: github.event_name == 'pull_request'` to the job so that it only runs if pull_request is going to have data in it.
> 
> What happened to this part? Don't we need to differentiate _pull_request_ and other events?

Hah. I apparently added it to the sambacc version instead of this PR:
https://github.com/samba-in-kubernetes/sambacc/pull/75/commits/dda4f72148d900c76cd67f27d8824534446398a1#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR15

Will fix.

hack/install-tools.sh
```diff
@@ -65,15 +73,19 @@ case "$1" in
 		_require_gobin
 		_install_gosec
 		;;
+	--gitlint)
+		_install_gitlint
+		;;
 	*)
 		echo "usage: GOBIN=<dir> $0 --<tool-name>"
 		echo ""
-		echo "availabel tools:"
```
obnoxxx, line 71: typo: availabel

obnoxxx (changes requested):
looks good in general but found one typo ...

obnoxxx (dismissed):
typofix mentioned previously

hack/install-tools.sh
obnoxxx: > This change fixes that typo

oh, now I see! thanks

phlogistonjohn: Thanks! I was starting to wonder if you were trolling me. ;-)

phlogistonjohn:
```
Storing signatures
kustomize not found in PATH, checking /root/samba-operator/.bin
make: *** [Makefile:256: kustomize] Error 1
Build step 'Execute shell' marked build as failure
Performing Post build task...
```
I don't see how my change would have broken the install for kustomize... but I did touch the file that makes it happen.
I tried running it locally and it succeeded.

obnoxxx (dismissed):
after more careful review, the PR lgtm, but a ci check is failing. this needs to be fixed to meet all requirements

anoopcs9:
> ```
> Storing signatures
> kustomize not found in PATH, checking /root/samba-operator/.bin
> make: *** [Makefile:256: kustomize] Error 1
> Build step 'Execute shell' marked build as failure
> Performing Post build task...
> ```
> 
> I don't see how my change would have broken the install for kustomize... but I did touch the file that makes it happen. I tried running it locally and it succeeded.

Even my local run completed successfully with the patch. I'll take a look at it later.

anoopcs9 (approved):

phlogistonjohn:
I fixed the CI failure.  install-tools.sh broke silently when either go (not in this case) or python3 (which failed in this case) were not on the system it also didn't matter if that was being used or not, so I updated the script with a 2nd patch.

anoopcs9 (approved):
lgtm.

anoopcs9:
> I fixed the CI failure. install-tools.sh broke silently when either go (not in this case) or python3 (which failed in this case) were not on the system it also didn't matter if that was being used or not, so I updated the script with a 2nd patch.

Awesome find 👏🏻 

Especially I liked the part where you added another commit instead of squashing the dependency change into previous one which kind of helped me easily understand the "missing python3" error in CentOS CI.

obnoxxx (approved):
lgtm

PR #307, opened 2023-05-17
go.mod: upgrade to newer versions

synarete:
Upgrade logr, client-go and controller-runtime to newer versions which include various bug fixes. Align code with respect to changes in those libraries.

depends on #310


phlogistonjohn:
Thanks!  This has been on my TODO list for a while so I'm happy to see this happen.
On quick skim most things look OK, but since it is a big commit I will need a bit more time to fully review.

phlogistonjohn:
Commit message nitpick: change `upgrade to newer versions` to `update packages to newer versions`. That way we're clarify that it's not a bump of the required go version, but impacts the dependencies.

internal/resources/util_fakes_test.go
```diff
@@ -34,6 +34,8 @@ type fakeClient struct {
 	clientGet func(context.Context, types.NamespacedName, rtclient.Object) error
 }
 
+type fakeSubResource struct{}
```
phlogistonjohn, line 37: Could you move this to directly above the methods? I think Go is OK with that and I was initially confused as to how fakeSubResource could meet the SubResourceClient interface until I scrolled much further down.

phlogistonjohn (changes requested):
A nitpick, nothing major

phlogistonjohn (approved):
LGTM. Never hesitate to ping me if I ask for something and then you don't hear back from me for a while. :-)

phlogistonjohn:
@obnoxxx  @spuiuk 2nd review please?

anoopcs9 (approved):

anoopcs9:
@Mergifyio rebase

PR #308, opened 2023-05-18
resources: refine get-create logic for openshift elems

synarete:
When doing reconcile logic of Get-or-Create for OpenShift's elements, treat AlreadyExists error as transient error and just Get the element once again from the API server. This may happen in cases where there is a skew between apimachinery and client-go versions[1].

[1] https://github.com/kubernetes/client-go/issues/89

### DIFF
--- a/internal/resources/metrics.go
+++ b/internal/resources/metrics.go
@@ -124,13 +124,19 @@ func (m *SmbShareManager) getOrCreateMetricsService(
 	}
 	err = m.client.Create(ctx, srvWant, &rtclient.CreateOptions{})
 	if err != nil {
-		if errors.IsAlreadyExists(err) {
-			m.logger.Info("Metrics Service already exists", "key", srvKey)
-		} else {
+		if !errors.IsAlreadyExists(err) {
 			m.logger.Error(err, "Failed to create metrics Service",
 				"key", srvKey)
+			return nil, false, err
 		}
-		return nil, false, err
+		m.logger.Info("Retry to get metrics Service", "key", srvKey)
+		srvCurr, srvKey, err = m.getMetricsServiceOf(ctx, pl, ns)
+		if err != nil {
+			m.logger.Error(err, "Failed to get existing metrics Service",
+				"key", srvKey)
+			return nil, false, err
+		}
+		return srvCurr, false, err
 	}
 	return srvWant, true, nil
 }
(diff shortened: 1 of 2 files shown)


phlogistonjohn (approved):
Code change looks OK.  Was this weird behavior with the client something you experienced?

synarete:
> Code change looks OK. Was this weird behavior with the client something you experienced?

YES! It is also the motivation behind https://github.com/samba-in-kubernetes/samba-operator/pull/307

internal/resources/metrics.go
```diff
@@ -124,13 +124,19 @@ func (m *SmbShareManager) getOrCreateMetricsService(
 	}
 	err = m.client.Create(ctx, srvWant, &rtclient.CreateOptions{})
 	if err != nil {
-		if errors.IsAlreadyExists(err) {
-			m.logger.Info("Metrics Service already exists", "key", srvKey)
-		} else {
+		if !errors.IsAlreadyExists(err) {
 			m.logger.Error(err, "Failed to create metrics Service",
 				"key", srvKey)
+			return nil, false, err
```
anoopcs9, line 130: IIUC the linked [issue](https://github.com/kubernetes/client-go/issues/89) discussed a different aspect where actual error returned is not understood/processed by `errors` package. It also mentions about `IsAlreadyExists` being never returned for an existing object.

For us the change here is to bypass `IsAlreadyExists` error once and hopefully retrieve next time if it got created. Is that correct?

synarete: @anoopcs9 This PR uses the following pattern:

1) Get a resource
2) If OK, return it
3) If `!errors.IsNotFound(err)` bail-out with err
4) Create the resource
5) If `!errors.IsAlreadyExists(err)` bail-out with err
6) Get the existing resource

The linked issue hints that steps 4-5-6 could happen in cases where there is version mismatch between client-go and k8s version. Admittedly, I don't fully understand the internal's of client-go to explain *why* this could happen.

anoopcs9: > @anoopcs9 This PR uses the following pattern:
> 
>     1. Get a resource
> 
>     2. If OK, return it
> 
>     3. If `!errors.IsNotFound(err)` bail-out with err

Ok. Above steps were not visible in the default `git diff` view and I had to manually expand to get the complete context.

anoopcs9 (approved):

anoopcs9:
@Mergifyio rebase

PR #313, opened 2023-07-31
Bump minimum Go version to 1.19

anoopcs9:
depends on https://github.com/samba-in-kubernetes/samba-centosci/pull/28

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
-          go-version: "1.18"
+          go-version: "1.19"
       - uses: actions/checkout@v3
       - name: Build
         run: make
@@ -23,7 +23,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
-          go-version: "1.18"
+          go-version: "1.19"
       - uses: actions/checkout@v3
       - name: Install revive
         run: go install github.com/mgechev/revive@latest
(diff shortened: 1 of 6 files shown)


anoopcs9:
Not urgent, we could wait till Go 1.21 is released(may be in 2-3 weeks time).

phlogistonjohn (dismissed):
I'm perpetually perplexed by the precise meaning of the go version present in the go.mod file. I'm pretty positive that they're even planning to change it. Usually I prefer to see the build versions change but not the mod file. But I am not going to pester you about it.

anoopcs9:
> Not urgent, we could wait till Go 1.21 is released(may be in 2-3 weeks time).

Go [1.21](https://go.dev/doc/go1.21) is out now.

obnoxxx (changes requested):
I think "build:" is not the appropriate componeb for both the commit and the PR since this PR changes both the build itself and the github workflow actions. 

I think these two changes should be separated in two commits and maybe two dependent PRs ...

phlogistonjohn (dismissed):
Approved.

I agree with @obnoxxx that the `build` commit probably should be two commits but it doesn't bother me enough to block the PR.

Unfortunately, we're now going to continue to play catch-up because Go 1.21 has been released, making 1.19 now out of support so we should follow up with updates to 1.20... at least for the CI and containers (hopefully not go.mod again)

anoopcs9:
> Approved.
> 
> I agree with @obnoxxx that the `build` commit probably should be two commits but it doesn't bother me enough to block the PR.

Updated once more.

phlogistonjohn (commented):
LGTM, thanks

phlogistonjohn (approved):
LGTM

phlogistonjohn:
@obnoxxx or @synarete -- a 2nd review please

obnoxxx (approved):

PR #288, opened 2023-02-23
hack: Bump golangci-lint to v1.51.2

anoopcs9:
Remove already [deprecated](https://golangci-lint.run/product/roadmap/#v1490) linter `varcheck`.

PR #314, opened 2023-08-15
workflows: use symbolic name in go-version

phlogistonjohn:
Depends on: #313 
Depends on: #288 

The workflow files that use Go directly (as opposed to inside a container) use actions/setup-go - this action gained an ability to use "relative versions" of `stable` and `oldstable` in the last year or two.

Switch our workflow file to use `oldstable` rather than a specific version number in order to keep up with releases from Go upstream. Use `oldstable` as to ensure the older but still actively maintained version of Go works.

This doesn't handle the container image... that's a WIP project for another PR.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
-          go-version: "1.19"
+          go-version: oldstable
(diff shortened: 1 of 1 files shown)


obnoxxx:
@mergifyio rebase

synarete (approved):
LGTM

anoopcs9 (approved):
That's cool.

PR #315, opened 2023-08-23
hack: add go-version-check.sh script

phlogistonjohn:
Related to, but not blocked on, #314 

Add a script that validates the Dockefile (or other specified container files) or shows the list of current supported golang version images. The goal is to ensure we are continually using a supported Go version and not falling behind.

Add a new check-dockerfile-go-version makefile target and include it in the check target. This runs the recently added script to validate the version of golang used to build in container.

The downside of this approach is that PRs & other work unrelated to Golang may periodically break when Go issues a new major release.
Let's try this for now and if it ends up getting really annoying we can try to do some other form of automation to alert us to the need of a version bump that doesn't interfere with unrelated work.

### DIFF
--- a/hack/go-version-check.sh
+++ b/hack/go-version-check.sh
@@ -1,0 +1,113 @@
+#!/bin/bash
+#
+# List or check that the golang version embedded in files (specifically our
+# Dockerfile) is using a supported version of Go.
+# Run with --show to list the images the script thinks are current.
+# Run with --check (the default) to verify the Dockerfile contains a current
+# image & version tag.
+#
+
+set -e
+SCRIPT_DIR="$(readlink -f "$(dirname "${0}")")"
+PROJECT_DIR="$(readlink -f "${SCRIPT_DIR}/..")"
+WORKDIR="$(mktemp -d)"
+CONTAINER_FILES=("${PROJECT_DIR}/Dockerfile")
+GOVERSURL='https://go.dev/dl/?mode=json'
+ACTION=check
+
+
+fetch_versions() {
+    local url="$1"
+    local out="$2"
+    if [ -f "${out}" ]; then
+        return
+    fi
+    echo "Fetching Go versions..." >&2
+    curl --fail -sL -o "${out}" "${url}"
+}
+
+extract_versions() {
+    local jsonfile="$1"
+    jq -r '.[].version' < "${jsonfile}" | \
+        cut -d. -f1,2 | sort -u | sed -e 's,^go,,'
+}
+
+to_images() {
+    for a in "$@"; do
+        echo "docker.io/golang:${a}"
+    done
+}
+
+cleanup() {
+    # only used in trap, ignoe unreachable error
+    # shellcheck disable=SC2317
+    rm -rf "${WORKDIR}"
+}
+trap cleanup EXIT
+
+
+opts=$(getopt --name "$0" \
+    -o "csf:" -l "check,show,file:,url:,versionsfile:" -- "$@")
+eval set -- "$opts"
+
+cli_cfiles=()
+while true ; do
+    case "$1" in
+        -c|--check)
+            ACTION=check
+            shift
+        ;;
+        -s|--show)
+            ACTION=show
+            shift
+        ;;
+        -f|--file)
+            cli_cfiles+=("$2")
+            shift 2
+        ;;
+        --url)
+            GOVERSURL="$2"
+            shift 2
+        ;;
+        --versionsfile)
+            GOVERSIONSFILE="$2"
+            shift 2
+        ;;
+        --)
+            shift
+            break
+        ;;
+        *)
+            echo "unexpected option: $1" >&2
+            exit 2
+        ;;
+    esac
+done
+
+if [ "${#cli_cfiles}" -gt 0 ]; then
+    CONTAINER_FILES=("${cli_cfiles[@]}")
+fi
+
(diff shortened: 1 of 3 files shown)


hack/go-version-check.sh
```diff
@@ -0,0 +1,111 @@
...
+# List or check that the golang version embedded in files (specifically our Dockerfile)
+# is using a supported version of Go.
+# Run with --show to list the images the script thinks are current.
+# Run with --check (the default) to verify the Dockerfile contains a current
+# image & version tag.
+#
+
+set -e
+SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+PROJECT_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
```
synarete, line 12: Consider change to:
```
PROJECT_DIR="$(readlink -f "${SCRIPT_DIR}/..")"
```

It avoids the extra call to `pwd`.

hack/go-version-check.sh
```diff
@@ -0,0 +1,111 @@
...
+WORKDIR="$(mktemp -d)"
```
synarete, line 13: If I read this correctly, in case user provides `GOVERSIONSFILE` this temp-directory is created and removed upon exit without being used -- is this correct?

hack/go-version-check.sh
```diff
@@ -0,0 +1,111 @@
...
+CONTAINER_FILES=("${PROJECT_DIR}/Dockerfile")
+GOVERSURL='https://go.dev/dl/?mode=json'
+ACTION=check
+
+
+fetch_versions() {
+    local url="$1"
```
synarete, line 20: Shell scripts typically use one of two intent style:
- TABs (AFAIK, the most common style in the UNIX/Linux world).
- 2-spaces ([Google style](https://google.github.io/styleguide/shellguide.html#indentation))

Using 4-space (which is Python style) is fine but less common.

hack/go-version-check.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
+#
+# List or check that the golang version embedded in files (specifically our Dockerfile)
```
synarete, line 3: Line longer than 80-chars

hack/go-version-check.sh
```diff
@@ -0,0 +1,111 @@
...
+    if [ -f "${out}" ]; then
+        return
+    fi
+    echo "Fetching Go versions..." >&2
+    curl --fail -sL -o "${out}" "${url}"
+}
+
+extract_versions() {
+    local jsonfile="$1"
+    jq -r '.[].version' < "${jsonfile}" | cut -d. -f1,2 | sort -u | sed -e 's,^go,,'
```
synarete, line 31: Long line, better split with:

```
jq -r '.[].version' < "${jsonfile}" | \
    cut -d. -f1,2 | sort -u | sed -e 's,^go,,'
```

synarete (commented):
OK, but few style comments

hack/go-version-check.sh
phlogistonjohn: Yes. I had considered trying to avoid that but it seemed like much more work for very little payoff because the downside of the extra mkdir/rmdir is negligible.

hack/go-version-check.sh
phlogistonjohn: 4-spaces 4-eva.  ;-)
I only tolerate tabs when it's mandated by the language or there's a tool to do canonical formatting for me. :-)

hack/go-version-check.sh
phlogistonjohn: done

hack/go-version-check.sh
phlogistonjohn: done

hack/go-version-check.sh
synarete: I agree -- better to keep it as is and have simple code.

hack/go-version-check.sh
```diff
@@ -0,0 +1,113 @@
...
+
+to_images() {
+    for a in "$@"; do
+        echo "docker.io/golang:${a}"
+    done
+}
+
+
+opts=$(getopt --name "$0" \
+    -o "csf:" -l "check,show,file:,url:,versionsfile:" -- "$@")
```
anoopcs9, line 43: Did you purposefully avoided the short command line(`-u`) for `--url` option?

hack/go-version-check.sh
```diff
@@ -0,0 +1,113 @@
...
+if [ "${#cli_cfiles}" -gt 0 ]; then
+    CONTAINER_FILES=("${cli_cfiles[@]}")
+fi
+
+cleanup() {
+    # only used in trap, ignoe unreachable error
+    # shellcheck disable=SC2317
+    rm -rf "${WORKDIR}"
+}
+trap cleanup EXIT
```
anoopcs9, line 89: To catch the exit for any unexpected option we need to place the `trap` well before those command-line parsing statements from above. Otherwise we end up with empty temporary directories.

anoopcs9 (changes requested):

hack/go-version-check.sh
phlogistonjohn: Right, done.

synarete (approved):
Good catch by @anoopcs9 -- LGTM now

anoopcs9 (approved):

PR #326, opened 2024-02-05
Explicitly restrict to amd64 hosts

tiferrei:
Until ARM support is ready, I have been running this version on my mixed cluster to make sure that only amd64 nodes serve the deployments.

PR #328, opened 2024-02-13
config: restrict deployments to amd64 nodes

synarete:
Until samba-operator supports other architectures (namely, ARM64), restrict its deployment to AMD64 nodes (e.g., in cases of non x86_64 or mixed ARCH clusters) via standard K8s node-selector mechanism [1].

[1] https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/

depends on #329

### DIFF
--- a/config/manager-full/auth_proxy_patch.yaml
+++ b/config/manager-full/auth_proxy_patch.yaml
@@ -8,6 +8,15 @@ metadata:
 spec:
   template:
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: kubernetes.io/arch
+                    operator: In
+                    values:
+                      - amd64
       containers:
         - name: kube-rbac-proxy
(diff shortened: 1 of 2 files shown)


obnoxxx (changes requested):
@synarete  could you please add a Co-authored tag instead of your signoff to the commit message? I. E. `Co-authored-by: Shachar Sharon <your email>` 

otherwise, this looks like a clean wrapping of PR #326

obnoxxx:
@phlogistonjohn  wrote.
> It might be nice to clarify that your patch is an adaptation of the original contribution. Something like adding a 'Original-Version-By:' footer?

Not sure it is required because  the original contributor @tiferrei  of #326  is still t author of this PR's  commit which also carries their signoff line

synarete:
> @synarete could you please add a Co-authored tag instead of your signoff to the commit message? I. E. `Co-authored-by: Shachar Sharon <your email>`
> 
> otherwise, this looks like a clean wrapping of PR #326

ACK

phlogistonjohn (approved):
good enough for me.

Issue #330, opened 2024-02-29
Samba-operator crashes when setting affinity in clustered samba server


FTS152:
Reproduce:

1. Open ctdb experimental feature
2. Set affinity options in smbcommonconfig
```
podSettings:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: kubernetes.io/hostname
            operator: In
            values:
    ...
```
3. Deploy smbshare and samba-operator crashes with the following logs:

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
 panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1593f0a]

goroutine 314 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
 /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.4/pkg/internal/controller/controller.go:119 +0x1fa
panic({0x16f5040, 0x26e0e20})
 /usr/local/go/src/runtime/panic.go:884 +0x213
github.com/samba-in-kubernetes/samba-operator/internal/resources.buildOneSmbdPerNodeAffinity(0xc0000caf00, 0xc000409bc0?, {0x18fc6c8, 0x20})
 /workspace/internal/resources/statefulsets.go:89 +0x22a
github.com/samba-in-kubernetes/samba-operator/internal/resources.buildStatefulSet(0xc0000caf00, {0xc0001470e0, 0x11}, {0xc0009f20e0, 0x1d}, {0xc000147098, 0x15})
 /workspace/internal/resources/statefulsets.go:34 +0x30d
github.com/samba-in-kubernetes/samba-operator/internal/resources.(*SmbShareManager).getOrCreateStatefulSet(0xc00081fc90, {0x1b47760, 0xc000409ad0}, 0xc0000caf00, {0xc000147098, 0x15})
 /workspace/internal/resources/getters.go:406 +0xc6
github.com/samba-in-kubernetes/samba-operator/internal/resources.(*SmbShareManager).updateClusteredState(0xc00081fc90, {0x1b47760, 0xc000409ad0}, 0xc0000caf00)
 /workspace/internal/resources/smbshare.go:351 +0xee
[...]

PR #331, opened 2024-02-29
resources: fix case of nil PodAntiAffinity

synarete:
Protect from case where user provided 'affinity' entry in podSettings of smbcommonconfig but without 'podAntiAffinity' sub-entry. This in turn would cause samba-operator a nil pointer dereference.

Fixes: https://github.com/samba-in-kubernetes/samba-operator/issues/330


anoopcs9 (approved):
Thanks for the quick fix.

PR #351, opened 2024-12-01
readme: add a maintenance status warning

phlogistonjohn:
Add a maintenance alert section to the readme. Ask for maintainers and contributors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,14 @@
 
 An operator for Samba as a service on PVCs in kubernetes.
 
+## ⚠️ Maintenance Status ⚠️
+
+Maintainers and contributors wanted!
+
+We wish to thank everyone who has shown interest in using and/or contributing to the samba-operator! However, we feel it important to share the fact that samba-operator is now minimally maintained. This means that the current team tries to keep up with basic "housekeeping" tasks but has little to no time for working on features or even some non-critial bug fixes. This is at least partly due to changing priorities for our employer - and since most of us work on this project as $DAYJOB task we don't have as much time to spend on this operator.
+
+We'd like this to be otherwise and as such we highly encourange anyone with a basic knowledge of Kubernetes, Go, or Operators who wants to see a healthy samba-operator consider getting involved. Feel free to contact the team via the [GitHub discsussions for the project](https://github.com/samba-in-kubernetes/samba-operator/discussions).
+
 ## Description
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,14 @@
 
 An operator for Samba as a service on PVCs in kubernetes.
 
+## ⚠️ Maintenance Status ⚠️
+
+Maintainers and contributors wanted!
+
+We wish to thank everyone who has shown interest in using and/or contributing to the samba-operator! However, we feel it important to share the fact that samba-operator is now minimally maintained. This means that the current team tries to keep up with basic "housekeeping" tasks but has little to no time for working on features or even some non-critial bug fixes. This is at least partly due to changing priorities for our employer - and since most of us work on this project as $DAYJOB task we don't have as much time to spend on this operator.
```
synarete, line 9: Missing 'is':
"...we feel it **is** important..."

synarete (approved):
LGTM (+minor comment)

README.md
phlogistonjohn: Serves me right for trying to do this quickly using the GH UI, now I need to figure out how to edit a commit. :-)

obnoxxx: @phlogistonjohn  qrote:
> Serves me right for trying to do this quickly using the GH UI,

Well, the obvious advantage of that is the immediate visual control of the ewndered document.

> now I need to figure out how to edit a commit. :-)

I don't know if it is possible in the UI at all.


Of course you know this, but you can do it locally with `git commit --amend`

obnoxxx: One alternative: If the change request is added as a suggested diff, you can apply it as an additional commitfrom the GH UI, but then it would need to be squashed with your original commit and I think that is not possible from the UI but has to be done locally with `git rebase -i` 

So as a summary, I think it is not possible yo do it entirely from the UI, only partially.

obnoxxx (commented):
Looks good in general, but  I agree to the missing `is` that  @synarete  spotted.

README.md
obnoxxx: @phlogistonjohn , we could also fix this in a follow-up PR :)

obnoxxx (approved):
Made up my mind: we can change that one thing in a follow-up PR. 
LGTM.

README.md
obnoxxx: see #352

PR #355, opened 2025-03-13
go.mod: Bump Go version to 1.23.0

anoopcs9:
It is really unfortunate that golang has decided to come up with an [automation](https://github.com/golang/go/issues/69095) to update `go` directive in _go.mod_ from various repositories including x/repos following the release cadence in general. Another interesting and visible change is to explicitly specify the minor version as a suffix to the major version. Thus we are bound to update our `go` directive as we require many of the golang x/repos.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM docker.io/golang:1.22 as builder
+FROM docker.io/golang:1.23 as builder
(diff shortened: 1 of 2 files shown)


phlogistonjohn (approved):
Go is a treadmill.

Change LGTM.

synarete (approved):
lgtm

Issue #281, opened 2023-02-13
Custom SMB configuration options


taxilian:
Is there a way to add custom SMB config options? For example I am trying to create a share that I can use for provisioning servers, but the management tools for the servers can only use the oldest SMB1 (NT1) protocol, so I need to override the "min version" in the config file.

Is there a place that I'm just not finding for adding things like that?

taxilian:
No worries -- it turns out the client is even worse than suspected, when I use it with a loadbalancer IP it won't connect at all. By comparing the difference with tcpdump it appears that it won't even attempt to connect to something that doesn't respond to a ping -- and when I tried connecting it to the pod directly it worked but only very poorly, which leads me to suspect it's not designed to work with anything not on the same local network (which my kube cluster sorta technically isn't since it requires BGP routing via the ToR router to get to).

... so, I had to give up on putting it in my cluster, sadly.

Thank you for the response!

taxilian:
Oh -- as a workaround for anyone in the future looking for this, though, you can edit the configmap after creating the resource and it will work. That will probably get recreated if you change the resource, but it does at least let you put in custom config options.

phlogistonjohn:
It's too bad about that ping thing. Too often people write things like that are "helpfully" unhelpful.

And because I feel obligated to mention it: you can still use our sibling project [samba-containers](https://github.com/samba-in-kubernetes/samba-container) Samba server images on standalone nodes. If it's easy enough in your env. to install samba from packages, no worries, but I do like to promote the container images when I can :-)

taxilian:
Yeah, that's a point -- probably should have done that. Maybe I could even figure out how to set up docker to use cephfs still =]

I may look into it, just ran out of time. That would at least give me an easily repeatable option. I appreciate the suggestion!

tiferrei:
I don't wish specifically to comment on an old(er) issue but I just wanted to report my use case that would be greatly improved by having custom global configs. 

I mostly connect to my shares on a Mac, which uses the built-in SMB stack. It is known for being _horrible_ however I guess it is just the one that I have access to. An issue I've been facing a lot recently is that whenever I try to copy a file into the share I get a "file in use error" that seems to be resolved by using some special `vfs_fruit` configs that make the experience easier on Mac.

[...]

PR #357, opened 2025-04-21
resources: add support for setting custom samba configuration

FTS152:
Add new fields to handle custom Samba configurations.

For [global] section, a new field `customGlobalConfig` is added under `smbCommonConfig` to handle custom settings. For individual share configurations, modifications can be made in `customShareConfig` field under smbShare. This design is chosen because, in group mode, a single server may correspond to several grouped shares.

For example, we can set `customGlobalConfig` in `smbCommonConfig` like:
```
apiVersion: samba-operator.samba.org/v1alpha1
kind: SmbCommonConfig
metadata:
  name: samba-service-mypublished
  namespace: samba-operator-system
spec:
  customGlobalConfig:
    configs:
      aio write size: "1"
      strict rename: "Yes"
  network:
    publish: external
```
During Update() process for configuration, the applyCustomGlobal() function is used to apply the keys and values from `smbCommonConfig.spec.customGlobalConfig.configs` into smb.conf, which will take effect on the deployed server. However, this method currently has some limitations. When samba server is already running, it is not possible to update smb.conf by modifying `smbCommonconfig` without interrupting the service. This is mainly due to the following constraints:

1. Updates to the ConfigMap are not immediately reflected inside the pod’s volume (e.g., /etc/container-config/config.json)
2. After modifying `smbCommonConfig`, you also need to import the changed config json to smbd manually (by execute samba-container import in the pod).
3.  The apply function currently only supports adding or modifying existing fields; it does not yet support field deletion.

These limitations are not insurmountable. The ideal solution would be to implement a mechanism that watches for changes to the config.json inside the pod, automatically imports the new configuration into smbd when changes are detected, and we should also improves the logic of the apply function. For now, this commit only provide the ability for modifying samba configurations when deploying a samba server.

related issue: #281


phlogistonjohn:
Thanks for the PR. FWIW this is what we implemented in ceph for something similar:
https://docs.ceph.com/en/latest/mgr/smb/ (search for `custom_smb_global_options` and/or `custom_smb_share_options`). It's similar but since the custom options really can break the automation I require a special key to check / hint that the user read the docs and understands the risks of setting a custom config option.

FTS152:
OK, I will add a key to check if the user already read docs and know the risks that custom configs may lead to unexpected behaviors. Custom settings are ignored if acknowledgement flags are not set:

```
apiVersion: samba-operator.samba.org/v1alpha1
kind: SmbCommonConfig
metadata:
  name: samba-service-mypublished
  namespace: samba-operator-system
spec:
  customGlobalConfig:
    useUnsafeCustomConfig: true    <-- add this explicitly
    configs:
      aio write size: "1"
      strict rename: "Yes"
  network:
    publish: external
```

api/v1alpha1/smbcommonconfig_types.go
```diff
@@ -38,6 +38,12 @@ type SmbCommonConfigSpec struct {
 	// under PodSettings allow admins and users to customize how pods
 	// are scheduled in a kubernetes cluster.
 	PodSettings *SmbCommonConfigPodSettings `json:"podSettings,omitempty"`
+
+	// GloabalConfig are configuration values that are applied to [gloabal]
```
s-burris, line 42:
```suggestion
	// GlobalConfig are configuration values that are applied to [global]
```

config/crd/bases/samba-operator.samba.org_smbcommonconfigs.yaml
```diff
@@ -30,6 +30,18 @@ spec:
             spec:
               description: SmbCommonConfigSpec values act as a template for properties of the services that will host shares.
               properties:
+                customGlobalConfig:
+                  description: GloabalConfig are configuration values that are applied to [gloabal] section in smb.conf for the smb server. This allows users to add or override default configurations.
```
s-burris, line 34:
```suggestion
                  description: GlobalConfig are configuration values that are applied to [global] section in smb.conf for the smb server. This allows users to add or override default configurations.
```

FTS152: Thanks for pointing out typos!

phlogistonjohn (commented):
the change generally looks ok to me. I would request that the `fix` commits get folded/squashed into the original patches to have a nice clean history. If that's ok but you want help doing so just ask.

FTS152:
Sure, I'd appreciate it if you could help with the commit squash. Thanks!

phlogistonjohn:
First execute `git checkout master && git pull --ff-only && git checkout feat_enable_custom_conf && git rebase master` to make sure your current branch is up to date with our base branch.

Then run `git rebase -i master` it will bring up an editor window that looks a bit like:
```
pick 6634d266f62 foo
pick 761c2c172aa bar
pick 0cdf01debf1 baz
pick 1d28814ffd6 quux
```
(these are made up ids and commits of course)

Edit this to move the fixup commits after the commit that added the comments with typos. I think in your case it's the first unique commit in the branch. Then change `pick` to `squash` for those two. It'll look something like:

```
pick 6634d266f62 foo
squash 0cdf01debf1 baz
squash 1d28814ffd6 quux
pick 761c2c172aa bar
```
Save and exit the editor.

When you squash the commits the tool will give you an opportunity to edit the commit message. Do so if you wish.
Look over your changes and if everything looks good do a force push. If you get stuck think you've messed up your commits let me know and I can take more direct measures but I prefer to try and coach you to do it yourself first :-)

FTS152:
Thanks for the clear and detailed explanation! Well...I think I've messed up something at the beginning (I included a commit in master branch) so I force-pushed twice. Please let me know if there are still any other problems.

anoopcs9:
> @anoopcs9. Thanks for getting this feature into master. Would it be possible to cut release with this functionality?

We are already due for a release. I hope we can get it out sooner rather than later. @phlogistonjohn any further comments?

HasseJohansen:
Ok. I could maybe be up for it, but I am definitely not a Go expert 🙂but I am willing to learn

phlogistonjohn:
Awesome. For now, I just suggest keeping on top of issues & PRs and providing feedback when possible. Also, let's use gihub discussions for general discussions, rather then keep commenting on this closed item :-)

PR #362, opened 2025-06-30
mergify: Restrict the addition of "priority-review" label

anoopcs9:
Mergify rules doesn't check the status of pull requests before adding the "priority-review" label. In addition to the existing rules it only makes sense to add the label if the pull request is in open state.


phlogistonjohn (approved):

obnoxxx (approved):
LGTM, thanks.

PR #365, opened 2025-09-23
Update basic tools and environment for go1.25

anoopcs9:
- [go1.25](https://go.dev/blog/go1.25) is now available making go1.24 the oldest supported release. Accordingly update the Dockerfile to make _go-version-check.sh_ happy.
- Bump `golangci-lint` and `gosec` to a version that supports go1.24 (and go1.25).
- Update `controller-gen` to support also go1.25.

PR #366, opened 2025-09-23
hack: Update controller-gen to v0.19.0

anoopcs9:
With go1.25, there's a [breaking change](https://github.com/golang/go/issues/74462) from _golang.org/x/tools_ pulled in by controller tools for which a fix is available in either v0.24.1 (and above) or v0.25.1 (and above). This minimum version of _golang.org/x/tools_ is only included from controller tools v0.16.4. We make use of this opportunity to fetch the latest version.

### DIFF
--- a/hack/install-tools.sh
+++ b/hack/install-tools.sh
@@ -36,7 +36,7 @@ _install_kustomize() {
 }
 
 _install_controller_gen() {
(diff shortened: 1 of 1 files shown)


anoopcs9:
Ah..`make check` failures need fixes from #365. Therefore closing the PR by including the change as part of #365.

PR #368, opened 2026-01-06
github: Pin versions for GitHub actions for dependabot automation

### LINKED ISSUE #367: Update github actions to use pinned versions
Following a similar policy in the ceph organization I'm encouraging all projects in this org to use pinned versions. That's expressing a specific hash versus a tag like so:
```
# "bad"
- uses: actions/checkout@v4
# good
- uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8
```

See also: https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions
We should apply this to *all* actions outside our direct control including the ones provided by github.

To keep such dependencies up to date, instead of using a tag, we can enable dependabot which will detect newer versions of an action and file PRs to suggest updating to said action. This PRs can be reviewed like any other PR.
See https://github.com/samba-in-kubernetes/samba-container/blob/master/.github/dependabot.yml for an example.

anoopcs9:
fixes #367

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "semiannually"
+    rebase-strategy: disabled


obnoxxx (approved):
I have not checked each and every hash but in general, this change looks good.

one point of criticism:

The PR description does not contain a meaningful description.

I request to write more detailed descriptions for future PRs.

I like the fact that the `dependabot.yml` file is added in a separate commit. I might have suggested to even have a separate PR for this but will not insist now.

LGTM

anoopcs9:
@Mergifyio queue

PR #371, opened 2026-02-21
buildSmbMetricsContainer: pass --port to smbmetrics container

BenEstrabaud:
This fix is about buildSmbMetricsContainer in internal/resources/metrics.go: when the smbmetrics sidecar container is enabled, the samba share pod always crashes (CrashLoopBackoff).

buildSmbMetricsContainer() configures liveness and readiness probes on port 8080 but does not pass --port to the smbmetrics binary, which defaults to port 9922. The probes therefore always fail, putting the samba-metrics sidecar into a CrashLoopBackOff immediately after startup.

Fixed by passing --port= as an argument so the binary listens on the same port the probes expect.

This fix was tested on a live k3s cluster:

```kubectl get pods -n samba -o custom-columns='NAME:.metadata.name,STATUS:.status.phase,CONTAINERS:.spec.containers[*].name'
NAME                           STATUS    CONTAINERS
storage-6695849ccc-vgcph       Running   samba,watch-update-config,samba-metrics
timemachine-6cb85c447d-7fs9f   Running   samba,watch-update-config,samba-metrics```

### DIFF
--- a/internal/resources/metrics_test.go
+++ b/internal/resources/metrics_test.go
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package resources
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestBuildSmbMetricsContainer(t *testing.T) {
+	ctr := buildSmbMetricsContainer("quay.io/samba.org/samba-metrics:latest", nil, nil)
+
+	t.Run("name", func(t *testing.T) {
+		assert.Equal(t, "samba-metrics", ctr.Name)
+	})
+
+	t.Run("image", func(t *testing.T) {
+		assert.Equal(t, "quay.io/samba.org/samba-metrics:latest", ctr.Image)
+	})
+
+	t.Run("portArgMatchesProbes", func(t *testing.T) {
+		// The --port arg passed to the binary must match the port used by the
+		// liveness and readiness probes, otherwise the probes always fail and
+		// the container enters a CrashLoopBackOff.
+		expectedArg := fmt.Sprintf("--port=%d", defaultMetricsPort)
+		assert.Contains(t, ctr.Args, expectedArg,
+			"smbmetrics binary must receive --port matching the probe port")
+
+		assert.Equal(t, int32(defaultMetricsPort),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: internal/resources/metrics_test.go [hunk 1]
```diff
@@ -0,0 +1,53 @@
...
+	ctr := buildSmbMetricsContainer("quay.io/samba.org/samba-metrics:latest", nil, nil)
+
+	t.Run("name", func(t *testing.T) {
+		assert.Equal(t, "samba-metrics", ctr.Name)
+	})
+
+	t.Run("image", func(t *testing.T) {
+		assert.Equal(t, "quay.io/samba.org/samba-metrics:latest", ctr.Image)
+	})
+
```
synarete, line 23: Using `t.Run()` for such checks is too much. A simple `assert.Equal` is good enough here.

synarete (commented):
The fix itself is in-place but the commit message does not follow the projects conversions:

1) Split long lines
2) Single white-line between commit title and body
3) Title should be shorter and correlate to actual change. Something like: "resources: pass '--port' option to smbmetrics container"

anoopcs9:
@BenEstrabaud While addressing the review comments, please ensure that the branch is rebased onto the latest master to incorporate the fix from #372.

BenEstrabaud:
No problem, I won't be able to do it for a few days but I'll incorporate all the suggestions when I get to it.

BenEstrabaud:
Alright I've updated my PR (I know, quite late, I was busy, sorry):
* Rebased on latest master
* Replaced t.Run with asserts
* Fixed the commit message.

Let me know if there still are some issues. I did notice that the `check-dockerfile-go-version` CI/CD check fails, but that's outside of the scope of my change, it fails because the Go version in Dockerfile is out of support since Go 1.27 released.

anoopcs9:
@Mergifyio rebase

anoopcs9:
> I did notice that the `check-dockerfile-go-version` CI/CD check fails, but that's outside of the scope of my change, it fails because the Go version in Dockerfile is out of support since Go 1.27 released.

This was fixed in the mean time (https://github.com/samba-in-kubernetes/samba-operator/pull/383) and the PR has been rebased.

anoopcs9 (approved):
lgtm, thanks.

synarete (approved):
lgtm

PR #375, opened 2026-03-28
resources: support PVC sub paths deeper than one level

twelho:
There does not seem to be a technical limitation for doing this, other than complexity involving path validation. This uses a defense-in-depth approach with an (admittedly somewhat complex) regex to prevent path escapes, coupled with a manual bounds check if the CR is applied without validation.


twelho:
Simplified the validation regex since negative lookahead is not supported. This is not as extensive, but should still serve its purpose.

PR #377, opened 2026-06-06
go.mod: update go version

daveol:
Update go version to align with the one in the Dockerfile, which was updated in #372 but not in go.mod

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/samba-in-kubernetes/samba-operator
 
-go 1.23.0
+go 1.25.11
 
 require (
 	github.com/go-logr/logr v1.2.3


daveol:
maybe the go version in `go.mod` should also be checked by `hack/go-version-check.sh` or a similar script?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/samba-in-kubernetes/samba-operator
 
-go 1.23.0
+go 1.25.11
```
anoopcs9, line 3: We try to stay at the minimum required Go version to compile and build the project. At the same time, we must raise it to match the highest version required by the dependencies listed under `require`. Since those dependencies haven't been updated for a while, `go mod tidy` doesn't raise any complaints. Until they are updated, something @dependabot should handle, and require a higher version, we prefer not to change it. This is independent of the CI version check for the Dockerfile.

daveol:
It seems this PR was unnecessary, sorry for the noise.

